PR #1, opened 2015-04-29
Fixed the handling of the return message from IO HTTP bind

bradtaylor:
We were not sending this message from within an actor, and previously nobody was handling the response. This was resulting in dead letters in the case of success, and provided no handling in the case of failure.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
@@ -2,11 +2,16 @@ package org.broadinstitute.dsde.agora.server
 
 import akka.actor.ActorSystem
 import akka.io.IO
+import akka.io.Tcp.CommandFailed
+import akka.pattern.ask
+import akka.util.Timeout
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.slf4j.LazyLogging
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.agora.server.webservice.ApiServiceActor
 import spray.can.Http
+import scala.concurrent.duration._
+import scala.concurrent.Await
 
 import scala.util.Try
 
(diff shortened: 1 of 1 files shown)


bradtaylor:
@jacarey @dshiga @geoffjentry @mcovarr

src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
```diff
@@ -32,8 +36,18 @@ class ServerInitializer(val config: Config) extends LazyLogging {
   }
 
   private def startWebServiceActors() {
+    implicit val timeout = Timeout(5.seconds)
```
nh13: why 5 seconds and why have it here?

src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
```diff
@@ -32,8 +36,18 @@ class ServerInitializer(val config: Config) extends LazyLogging {
...
+    import scala.concurrent.ExecutionContext.Implicits.global
     val service = actorSystem.actorOf(ApiServiceActor.props, "agora-actor")
-    IO(Http) ! Http.Bind(service, interface = webserviceInterface, port = webservicePort)
+    (IO(Http) ? Http.Bind(service, interface = webserviceInterface, port = webservicePort)).onComplete {
+      case Success(Http.CommandFailed(failure)) =>
```
geoffjentry: My only comment is about the Success(failure) type of pattern here. Until we start getting used to that as a group I'd either comment that this is intentional and why it happens or set up some structure to split the two apart

src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
bradtaylor: The ask requires it so I figured I'd keep it as close to this usage as possible. should it be up at the beginning of the class?

5 seconds because that's what Doug had. Better value?

nh13: I am not too worried about it.  Perhaps have it be a parameter to the method with a default value.

src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
```diff
@@ -32,8 +36,18 @@ class ServerInitializer(val config: Config) extends LazyLogging {
...
+        throw new Exception("could not bind to port: " + failure.toString)
+        actorSystem.shutdown()
+      case Failure(t) =>
+        throw new Exception("could not bind to port: " + t.toString)
+        actorSystem.shutdown()
+      case _ =>
+    }
```
mcovarr: Per conversations with Jeff and Doug yesterday, I think I would still lean toward the synchronous `Await.result` way Zamboni does this (not necessarily something I'll say often :smile: )

I believe this code currently works because the actor system is currently the only daemon thread. As soon as the main thread and this daemon thread exit, the application can exit.  But as our applications mature and become more complex, we might have more daemon threads launched in which case this would only take down the actor system but could leave the app running but unresponsive to web requests.

Furthermore if the bind fails the actor system thread might get shut down quickly, but the main thread would continue with pointless but potentially expensive initialization (I'm thinking of a Hibernate SessionFactory as an example here).  So even with a single daemon thread, instead of seeing a fast failure if the bind fails we'd have to wait for all the initialization to run.

src/main/scala/org/broadinstitute/dsde/agora/server/ServerInitializer.scala
bradtaylor: Needs to be an implicit. Going to keep it scoped close to the actual ask so it doesn't get passed around places we don't expect

PR #3, opened 2015-04-30
Added entity/metadata model as per our discussions today.

jacarey:
Basic data access layer added.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
@@ -1,0 +1,31 @@
+
+package org.broadinstitute.dsde.agora.server.dataaccess.mongo
+
+import com.mongodb.casbah.Imports._
+import com.mongodb.casbah.MongoCollection
+import com.mongodb.casbah.query.Imports
+import com.novus.salat._
+import com.novus.salat.global._
+import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
+
+  def findPayloadByRegex(regex: String): Seq[AgoraEntity] = {
+    find("payload" $regex regex)
+  }
+
+  def findByName(name: String): Seq[AgoraEntity] = {
+    find("metadata.name" $eq name)
+  }
+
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntityModel.scala
```diff
@@ -0,0 +1,9 @@
+
+package org.broadinstitute.dsde.agora.server.model
+
+import java.util.Date
+
+case class AgoraMetadata(id: Option[Integer] = None, namespace: String, name: String, synopsis: String,
```
geoffjentry: I get why you ordered this the way you did but putting defaulted params before non-defaulted params can lead to pain & anguish

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  def insert(entity: AgoraEntity): AgoraEntity = {
+    val entityDbObject = grater[AgoraEntity].asDBObject(entity)
+    collection.insert(entityDbObject)
+    find(entityDbObject).head
+  }
+
+  def find(query: Imports.DBObject) = {
+    (for (entity <- collection.find(query)) yield grater[AgoraEntity].asObject(entity)).toList
```
geoffjentry: If there's not a particular reason to use List (perhaps there is?) use Vector. It's way mo bettah

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
+
+  def findPayloadByRegex(regex: String): Seq[AgoraEntity] = {
+    find("payload" $regex regex)
+  }
+
+  def findByName(name: String): Seq[AgoraEntity] = {
```
bradtaylor: Will we want to do a find by name/namespace?

Also, for the present story, do we want findById?

jacarey: Both added in the next PR.

bradtaylor:
Looks great to me, taking into account Jeff's changes

PR #4, opened 2015-04-30
Fix travis badge in README

dshiga:
Should be travis-ci.org, not travis-ci.com.


dshiga:
some more changes to add

PR #7, opened 2015-05-01
Fix travis badge and coveralls

dshiga:
-Fixed travis badge in README.
-Updated scoverage and coveralls jars (only updated scoverage to 1.0.4 instead of 1.1.0 because 1.1.0 changed the way it looks for our source files and is breaking for obscure reasons)
-Removed some unused stuff in build.sbt (we don't need the key related stuff in there because we're not using travis pro and our coveralls key is in a travis environment variable, see https://github.com/scoverage/sbt-coveralls)

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,1 +1,1 @@
-service_name: travis-pro
+service_name: travis-ci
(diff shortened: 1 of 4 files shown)


dshiga:
By the way, the fix for coveralls was that travis needs a coveralls key in order to post the results to coveralls. It doesn't matter that agora is public (that only affects whether we need a *travis* key).

jacarey:
awesome! Merge it!

PR #14, opened 2015-05-01
Log to std out, a file and syslog. (the shotgun approach)

jacarey:


jacarey:
Shamelessly yanked from the vault-api

PR #16, opened 2015-05-02
Changed AgoraDao to find by AgoraEntity

jacarey:
Removed un-useful find methods and implemented a find: Seq and a findSingle: entity
Changed tests to use these methods. Tests currently broken.
Flattened data model
Merge changes.
Wire GET request to real db query.
Added custom marshall/unmarshall for AgoraEntity
Fix test cases.


jacarey:
Rebased and ready to rock

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -64,10 +62,22 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
...
+  }
+
+  override def find(entity: AgoraEntity): Seq[AgoraEntity] = find(EntityToMongoDbObject(entity))
+
+  override def findSingle(entity: AgoraEntity): AgoraEntity = {
+    val entityVector = find(EntityToMongoDbObject(entity))
+    entityVector.length match {
+      case 1 => entityVector.head
+      case 0 => throw new Exception("Found 0 documents matching: " + entity.toString)
+      case x if x > 1 => throw new Exception("Found >1 documents matching: " + entity.toString)
```
mcovarr: don't think you need the `if` :smile:

bradtaylor: What if you get -1? Probably that can't happen and we can just use _.

But what if mongo uses negatives as an error code or something? In that case, this match is not exhaustive

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  implicit val AgoraEntityMarshaller =
+    Marshaller.of[AgoraEntity](`application/json`) { (value, contentType, context) =>
+      context.marshalTo(HttpEntity(contentType, grater[AgoraEntity].toCompactJSON(value)))
+    }
+}
+
+@ApiModel(value = "Methods Query Response")
+case class AgoraEntity(@(ApiModelProperty@field)(required = true, value = "The method id")
+                       var id: Option[Int] = None,
```
bradtaylor: Should ID be versionId or something? I know we don't plan exposing the globally unique hash id to users, but could it cause confusion amongst devs, overloading "id"?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsServiceHandlers.scala
```diff
@@ -1,25 +1,29 @@
...
+      query(requestContext, namespace, name, id)
       context.stop(self)
   }
-  def query(requestContext: RequestContext, id: String): Unit = {
 
-    // TODO- Business logic to retrieve a method from the storage implementation
-    context.parent ! MethodsQueryResponse(id, Map("cwlFile" -> "fake/file/url"), Map("author" -> "Dave Shiga", "version" -> "1.0"))
+  def query(requestContext: RequestContext, namespace: String, name: String, id: Int): Unit = {
+    val method = AgoraDao.createAgoraDao.findSingle(namespace, name, id)
+    context.parent ! grater[AgoraEntity].toJSON(method)
```
bradtaylor: We still have to figure out marshalling in the perRequest, yah?

Could we just have complete() in that class expect from everyone, keeping it general, and delegate serialization to the service handler actors as we do here?

PR #17, opened 2015-05-04
Implement post endpoint

dshiga:


bradtaylor:
Looks good to me!

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -39,7 +39,10 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
     //insert the entity
     val dbEntityToInsert = EntityToMongoDbObject(entity)
     collection.insert(dbEntityToInsert)
-    findSingle(entity)
+    findSingle(entity) match {
```
mcovarr: this could be `getOrElse { throw new Exception(...) }`

PR #18, opened 2015-05-04
Add Request elements should not be optional.

bradtaylor:
We created a separate data model for POST requests so users could not supply values for things we want to create ourselves (like ID).
We also don't want them able to OMIT things we require. Like name. Therefore removed Options from these fields.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
@@ -26,13 +26,14 @@ object AgoraEntity {
     }
 
   def fromAgoraAddRequest(agoraAddRequest: AgoraAddRequest, createDate: Option[Date]) = {
-    new AgoraEntity(namespace = agoraAddRequest.namespace,
-                    name = agoraAddRequest.name,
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -71,15 +71,15 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = true, value = "The n
...
+                           name: String,
+                           @(ApiModelProperty@field)(required = true, value = "A short description of the method")
+                           synopsis: String,
+                           @(ApiModelProperty@field)(required = true, value = "Method documentation")
+                           documentation: String,
+                           @(ApiModelProperty@field)(required = true, value = "User who owns this method in the methods repo")
+                           owner: String, // TODO: remove (use authenticated user)
+                           @(ApiModelProperty@field)(required = true, value = "The method payload")
+                           payload: String
+                            )
```
dshiga: Does this mean all fields are required in an AgoraAddRequest?

bradtaylor: Yes, though user could supply an empty string (ie "")

PR #19, opened 2015-05-04
Add some simple validation

dshiga:
Uses the wdl parser to check payload for syntax errors when posting. Not sure we should merge yet because we're going to be getting much more thorough validation capabilities soon from the cromwell repo.


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsService.scala
```diff
@@ -1,9 +1,12 @@
 package org.broadinstitute.dsde.agora.server.webservice.methods
 
+import java.util.Scanner
```
bradtaylor: Are Scanner and WdlParser used in this class?

dshiga: Ah, good catch. I'll remove those imports.

bradtaylor:
This looks awesome.

.travis.yml
```diff
@@ -3,5 +3,7 @@ scala:
   - 2.11.6
 services:
   - mongodb
+jdk:
+  - oraclejdk8
```
bradtaylor: Does the WDL parser require java 8?  We are pulling java 7 into the docker image.

dshiga: It does in its current form. We *could* make it work with 7 but that adds some complications.

PR #20, opened 2015-05-05
Added an initial implementation of a CLI program.

bradtaylor:
Note- this works when tested against the CI deployment. See:
https://agora-ci.broadinstitute.org:443/methods/devtest/sample/1 through 5
which were all uploaded using this program.

Currently handles only methods push. Ultimately will want fetch as well. Also support for task-configurations.

Method push mimics git commit. Brings up a text editor to get user input for the synopsis/documentation metadata.

This is a rough initial implementation. Some extant issues:
We are currently assuming you have gotten the auth token as a string, and are supplying this on command line (this is what was done for apollo-testing). To make this really usable we need to enable programatic login to the OpenAm server to grab the token.

I'm not happy with the arg parsing/validation here. Grabs some defaults from a .apolloconfig, assumed to be in the same directory. Also meant to sort of mimic git. This is mostly for convenience, to save specifying from specifying user + namespace every single time. Also to set the target url (for CI vs Prod vs Dev deployments). But the validation here is a little ugly. Also, I'd like to enable multiple 'modes', push vs fetch, that carry a different set of required arguments. Not sure how to do this, but I'd love input.

### DIFF
--- a/src/scripts/agora.py
+++ b/src/scripts/agora.py
@@ -1,0 +1,164 @@
+################################################################################
+# agora.py
+# 
+# A CLI providing convenient access to the Agora methods repository REST API.
+#
+# author: Bradt
+# contact: dsde-engineering@broadinstitute.org
+# 2015
+# 
+# NOTES:
+# This is a very rough initial implementation. Needs a lot of work.
+#
+# Currently just supports method push. Ulitmately we will want method fetch, and
+#   task-configuration push/fetch.
+#
+# Must find a way to programatically get an Oath token from the openAM server.
+#   Currently, assumes you just grabbed the token and are passing it in as an
+#   argument. Really hackish.
+#
+# Gets default parameters from a .agoraconfig file, assumed to be in the same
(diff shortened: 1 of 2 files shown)


src/scripts/agora.py
```diff
@@ -0,0 +1,164 @@
...
+namespaceConfig = "namespace="
+urlConfig = "agora_url="
+
+def fail(message):
+    print message
+    sys.exit(1)
+
+
+# Read defualt values out of the configuration file. In particular, owner.
+# TODO - is there a smarter way to do this?
```
dshiga: You could do this with ConfigParser: https://docs.python.org/2/library/configparser.html

dshiga:
This is cool. The text editor thing will be handy. I could also imagine users wanting a way to do something like:
agora add test.wdl test_metadata.txt

where test_metadata.txt would contain:

namespace=broad
name=test
synopsis=This is a test method
documentation=This is a bunch of documentation.
  And this is the second line of documentation.

Not saying this needs to be supported now, but something to think about as a future enhancement.

PR #22, opened 2015-05-05
Modified dockerfile to bump java 7 -> java 8

bradtaylor:
To accomodate WDL parser, need to move from java7 to java8. There is no open-jdk java 8 package. Had to go Oracle.


dshiga:
I'll merge as soon as the builds pass

PR #23, opened 2015-05-06
Added test suite for bringing up agora and embedded mongo before all tes...

jacarey:
...ts and tearing it down after all tests.

Removing mongodb service from travis yaml.


src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+  override def beforeAll() {
+    println("Starting embedded mongo db instance.")
+    mongoProps = mongoStart(port = Agora.server.mongoDbPort)
+    println("Starting Agora web services.")
+    Agora.start()
+  }
+
+  override def afterAll() {
+    println("Stopping embedded mongo db instance.")
+    mongoStop(mongoProps)
```
bradtaylor: Agora.stop()? This was previously called at ApiTestFixture:26

bradtaylor:
One comment. Otherwise looks great!

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestSuite.scala
jacarey: Agora has a shutdown hook for stopping so it was a bit redundant. We can add it back though if you like the explicit stop call. You get duplicate stop messages in the logs though which is a bit odd.

PR #24, opened 2015-05-06
Removing mongodb service from dockerfile since we are using embeded mong...

jacarey:
...o.


bradtaylor:
if docker build works this should be good to go :)

PR #25, opened 2015-05-08
Initial implementation of search

dshiga:


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsService.scala
```diff
@@ -32,7 +32,36 @@ trait MethodsService extends HttpService with PerRequestCreator {
...
-          perRequest(requestContext, methodsQueryHandlerProps, ServiceMessages.Query(requestContext, namespace, name, id.toInt))
+          perRequest(requestContext, methodsQueryHandlerProps, ServiceMessages.QueryByNamespaceNameId(requestContext, namespace, name, id.toInt))
+      }
+    }
+
+  @ApiOperation(value = "Query for methods in the method repository",
+    nickname = "methods",
+    httpMethod = "GET",
+    produces = "application/json",
+    response = classOf[AgoraEntity],
```
bradtaylor: should be classOf[Seq]AgoraEntity]], yes?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -69,6 +88,23 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = true, value = "The n
                        payload: Option[String] = None
                         )
 
+@ApiModel(value = "Agora Search")
```
bradtaylor: Why do we need a separate AgoraSearch case class here? Just to avoid conflicts with the AgoraEntity companion object?

The alternative would be to use AgoraEntity and rename the companion object to be AgoraEntityMarshallers or similar, and then import it explicitly wherever (un)marshallers are needed.

Interested to here @jacarey's opinion

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -20,6 +20,16 @@ object AgoraEntity {
       case HttpEntity.Empty ⇒ new AgoraEntity()
     }
 
+  implicit val SeqAgoraEntityUnmarshaller =
+    Unmarshaller[Seq[AgoraEntity]](`application/json`) {
+      case HttpEntity.NonEmpty(contentType, data) => {
+        grater[AgoraEntity].fromJSONArray(data.asString)
+      }
+      case HttpEntity.Empty => {
+        Seq(new AgoraEntity())
```
jacarey: We may want to return an empty Seq... additionally I'm not sure returning new AgoraEntity above is a good idea either. Any thoughts?

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -29,6 +29,13 @@ class ApiServiceSpec extends FlatSpec with Matchers with Directives with Scalate
     }
   }
 
+  "Agora" should "return methods matching the query" in {
+    agoraDao.insert(testEntity2)
+    Get(ApiUtil.Methods.withLeadingSlash + "?namespace=broad") ~> methodsService.queryByEntityRoute ~> check {
```
bradtaylor: Do we want to test search with more than one parameter?
Probably namespace and owner (maybe want to remove the spaces from testMethod owner

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: Is there any reason we can't just always return a Seq of AgoraEntities? In which case we'd only need one marshaller/unmarshaller.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsQueryHandler.scala
```diff
@@ -30,4 +32,9 @@ class MethodsQueryHandler extends Actor {
       case Some(method) => context.parent ! grater[AgoraEntity].toJSON(method)
     }
   }
+
+  def query(requestContext: RequestContext, agoraSearch: AgoraSearch): Unit = {
+    val entities = AgoraDao.createAgoraDao.find(AgoraEntity.fromAgoraSearch(agoraSearch))
```
jacarey: As a side note I would be curious to stress test this. We need to consider that creating a new agora dao each time may not be the best practice. (Need to do some research in to mongo best practices)

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsService.scala
```diff
@@ -32,7 +32,36 @@ trait MethodsService extends HttpService with PerRequestCreator {
...
+    new ApiImplicitParam(name = "payload", required = true, dataType = "string", paramType = "path", value = "Payload")
+  ))
+  @ApiResponses(Array(
+    new ApiResponse(code = 200, message = "Successful Request"),
+    new ApiResponse(code = 500, message = "Internal Error")
+  ))
+  def queryByEntityRoute =
+    path(ApiUtil.Methods.path) {
+      get {
+        parameters('namespace?, 'name?, 'id.as[Int]?, 'synopsis?, 'documentation?, 'owner?, 'payload?).as(AgoraSearch) { agoraSearch =>
```
jacarey: That's cute.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
bradtaylor: What is the alternative? None doesn't seem to work as you are going HttpEntity => AgoraEntity

Seems to me, naively, that if you tried to unmarshall an empty HttpEntity to an AgoraEntity getting an empty one kind of makes sense

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
bradtaylor: This is used as the input to search, yes? We are returning Seq[AgoraEntity]

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: Switching everything to a Seq and returning an emtpy Seq should be fine I think. I'm not sure how marshalling would handle Options but I assume you could also change this to Option[AgoraEntity] and it would serialize the None case as well.

jacarey:
Looks good!

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
dshiga: 1) AgoraSearch doesn't have a createDate field, which helps me avoid complications in MethodsService.
2) Later we might want to have beforeDate and afterDate fields so people can specify a date range. But AgoraEntity has just one date field.

I did this mainly because of 1)

bradtaylor: Ahhh, the date issue. That makes sense. In that case I'm onboard.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
dshiga: But then would it be weird to have a findSingle method that returns a Seq[AgoraEntity]?

jacarey: I think findSingle can still return an Option[AgoraEntity] it is just that the query method would have to convert that in to a singleton Seq[AgoraEntity] before marshalling it.

I'm all for less marshallers/unmarshallers and keeping our model simple, but I'm open to any suggestions.

jacarey: Alternately we might be able to use delegate marshalling to use the same marshall for both a single AgoraEntity and a Seq[AgoraEntity]. That might be cleaner but I've never tried it.

dshiga:
I removed the unmarshallers (and marshaller) altogether, as they were only being used in tests. In the tests I'm now using salat grater directly where needed. Does that seem good, or do we think we'll need those unmarshallers for other stuff later?

PR #26, opened 2015-05-08
Bt documentation test

bradtaylor:
Added a test for successful storage of markdown in documentation
    
To satisfy DSDEEPB-46. Acceptance Criteria:

```
* There exists a metadata field which can accept a large (10KB) markdown representing tool documentation.
* There exists a unit test that validates that a 10kb string of github markdown is recovered without getting garbled
```

The metadata field exists. Here I'm adding the test. I made a little md file containing various markdown features.

I am concatenating that until it would be >10kb stored, and sending that through the webservice.
    
I am doing this the full way through the service, not just as a db test, because I want to make sure serialization/deserialization and spray marshalling etc. do not garble the string.

Only changed total testing time from 5s -> 6s

Also added Agora.stop() back to test suite cleanup.

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
+++ b/src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
@@ -3,13 +3,20 @@ package org.broadinstitute.dsde.agora.server
 import org.broadinstitute.dsde.agora.server.model.{AgoraAddRequest, AgoraEntity}
 
 trait AgoraTestData {
-
+  def getBigDocumentation: String = {
+    // Read contents of a test markdown file into a single string.
+    val markdown = io.Source.fromFile("src/test/resources/TESTMARKDOWN.md").getLines() mkString "\n"
+    markdown * 7  // NB: File is 1.6 Kb, so 7* that is >10kb, our minimal required storage amount.
+  }
+  
   val namespace1 = "broad"
   val namespace2 = "hellbender"
   val name1 = "testMethod1"
   val name2 = "testMethod2"
   val synopsis = "This is a test method"
   val documentation = "This is the documentation"
+  // NB: save io by storing output.
+  val bigDocumentation: String = getBigDocumentation
   val owner1 = "bob"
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -11,6 +13,17 @@ trait AgoraTestData {
 
   val documentation = "This is the documentation"
 
+  def getbigDocumentation: String = {
```
jacarey: B should be capitalized

geoffjentry: amusing that a little b was used in the word big

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -50,4 +50,11 @@ class ApiServiceSpec extends FlatSpec with Matchers with Directives with Scalate
     }
   }
 
+  "Agora" should "store 10kb of github markdown as method documentation and return it without alteration" in {
+    Post(ApiUtil.Methods.withLeadingSlash, marshal(testAddRequest)) ~> methodsService.postRoute ~> check {
+      val response = responseAs[AgoraEntity]
```
jacarey: any reason to store response as val?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -11,6 +13,17 @@ trait AgoraTestData {
...
+    val readMeFile = new File("src/test/resources/", "TESTMARKDOWN.md")
```
jacarey: do we need readMeFile? I don't think you use it. (Can remove java.io.File import too I think)

bradtaylor: Meant to clean this up. Will fix

PR #28, opened 2015-05-08
Fix search and unit tests

dshiga:
Searches on owner, documentation, synopsis, and payload were failing. I changed the order of fields in AgoraSearch to match those in AgoraEntity and it is now working.

I also fixed the unit tests in ApiServiceSpec. They were erroneously passing because we had statements using === but not wrapped in assert(). So even if the === statements return false, the tests would pass.

Lastly, I added additional unit tests to verify search on owner, documentation, synopsis, and payload.


jacarey:
Looks good. I'm going to merge this so I can start the rebase of the refactor pull request which is bound to get hairy

PR #29, opened 2015-05-09
Removed salat for bson -> json and back. Just use JSON mongo util.

jacarey:
Removed salat for bson -> json and back. Just use JSON mongo util. Updated per request from Rawls. String to DateTime implicit conversion to make the search work on a standard AgoraEntity. Added AgoraApiJsonSupport back since we
 are using spray json again.

We probably want to make sure everyone has their code merged back to master before we merge this guy.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/PerRequest.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/PerRequest.scala
@@ -1,19 +1,18 @@
 package org.broadinstitute.dsde.agora.server.webservice
 
-import akka.actor._
 import akka.actor.SupervisorStrategy.Stop
-import PerRequest._
+import akka.actor.{OneForOneStrategy, _}
+import org.broadinstitute.dsde.agora.server.webservice.PerRequest._
+import spray.http.HttpHeader
 import spray.http.StatusCodes._
+import spray.httpx.marshalling.ToResponseMarshaller
 import spray.routing.RequestContext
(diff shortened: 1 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/PerRequest.scala
```diff
@@ -36,32 +32,66 @@ trait PerRequest extends Actor with Json4sSupport {
...
+      complete(InternalServerError)
   }
 
-  def complete[T <: AnyRef](status: StatusCode, obj: T) = {
-    r.complete(status, obj)
+  /**
+   * Complete the request sending the given response and status code
+   * @param response to send to the caller
+   * @param marshaller to use for marshalling the response
+   * @tparam T the type of the response
```
mcovarr: @tparam is not a Javadoc thing

jacarey:
Ok I've rebased master with the changes @dshiga made to tests.

dshiga:
How are we validating required fields now? e.g. how are we preventing someone from adding a method without a name?

jacarey:
We will need to make per call validations so that we don't need a bunch of duplicated entities just for the sake of validation. I can add that to this branch.

dshiga:
I'm ok with merging this as is as long as we put it on our list of to do's.

jacarey:
Case class validation is nice for global validations across all endpoints.. sadly that doesn't really help us much so I think we need to do validate directives in our spray routing.

PR #30, opened 2015-05-12
Changed 'id' to 'snapshotId'

gbggrant:


src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -15,7 +15,7 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 import spray.json._
 
 object AgoraMongoDao {
-  val MongoDbIdField = "_id"
+  val MongoDbIdField = "_snapshotId"
```
dshiga: Do we want to keep this one as _id? We really just want the field name in the methods entity to be called snapshotId. Here we're changing a field name in the counter collection.

gbggrant: Thanks for catching that.

dshiga:
It looks good to me, but it might be good to get one more opinion.

jacarey:
+1

PR #31, opened 2015-05-12
Make json returned from search more concise

dshiga:
Also, fixed ApiServiceSpec tests so they are capable of failing.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
@@ -65,11 +65,14 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
     currentCount.get(CounterSequenceField).asInstanceOf[Int]
   }
 
-  def find(query: Imports.DBObject) = {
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -65,11 +65,14 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
...
   }
 
-  def find(query: Imports.DBObject) = {
-    (for (dbObject <- collection.find(query)) yield MongoDbObjectToEntity(dbObject)).toVector
+  def find(query: Imports.DBObject, projection: Option[Imports.DBObject] = None) = {
+    (for (dbObject <- collection.find(query, projection.getOrElse(MongoDBObject()))) yield MongoDbObjectToEntity(dbObject)).toVector
   }
 
-  override def find(entity: AgoraEntity): Seq[AgoraEntity] = find(EntityToMongoDbObject(entity))
+  override def find(entity: AgoraEntity): Seq[AgoraEntity] = {
```
jacarey: We might want to add a find that allows you to specify the fields to be projected in a generic way for the dao trait. We could then construct the projection for mongo in the mongo specific dao. Not sure exactly how we would implement that but we should discuss it.

jacarey:
Looks good.

PR #33, opened 2015-05-14
Changed Post Request to return 201 (rather than 200) when successfully adding a method

gbggrant:
Changed Post Request to return 201 (rather than 200) when successfully adding a method
Updated unit test to verify this.
Added a bit more testing of 200 codes being returned elsewhere.


src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -73,6 +77,7 @@ with AgoraTestData with AgoraDbTest {
       assert(agoraEntity.payload === payload1)
       assert(agoraEntity.snapshotId !== None)
       assert(agoraEntity.createDate != null)
+      assert(status === Created)
```
jacarey: I would add the phrase "with a status 201" to the test description so that it is explicit.  For instance "Agora" should "create and return a method with a status of 201"

jacarey:
:+1:

PR #34, opened 2015-05-14
Added cromwell and vault common from artifactory.

jacarey:
Added mock OpenAM so we can run unit tests without integration testing.
Pull owner out of the OpenAM ticket on creation.

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
@@ -1,12 +1,12 @@
 package org.broadinstitute.dsde.agora.server.webservice
 
+import org.broadinstitute.dsde.agora.server.AgoraTestData
 import org.broadinstitute.dsde.agora.server.business.AgoraBusiness
 import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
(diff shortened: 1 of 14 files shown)


gbggrant:
Looks good.  :+1:

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -41,63 +40,86 @@ with AgoraTestData with AgoraDbTest {
...
+      handleError(
+        entity.as[AgoraEntity],
+        (entity: AgoraEntity) => assert(entity.documentation.get === bigDocumentation)
+      )
     }
   }
 
-  def handleDeserializationErrors[T](deserialized: Deserialized[T]): T = {
-    if (deserialized.isRight) deserialized.right.get else failTest(deserialized.left.get.toString)
+  def handleError[T](deserialized: Deserialized[T], assertions: (T) => Unit) = {
```
dshiga: Assertion failures seem to be ignored when wrapped in handleError(). For example if I try adding assert(false === true) into the tests above, they still pass.

jacarey: Good catch. I never passed the right side of the either to the assertion function.

PR #35, opened 2015-05-14
Add URL for detailed method record to each search listing

dshiga:
I also made changes to how we're handling configuration, to avoid having duplicate lines like the following in multiple classes:
lazy val port = appConfig.as[Option[Int]]("webservice.port").getOrElse(8000)

I moved lines like this into AgoraConfig itself so they happen in one central place.

Most of the changes here are due to this configuration refactoring.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -1,0 +1,36 @@
+package org.broadinstitute.dsde.agora.server.business
+
+import org.broadinstitute.dsde.agora.server.AgoraConfig
+import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+/**
+ * Created by dshiga on 5/13/15.
+ */
+object AgoraBusiness {
+
+  def agoraUrl(entity: AgoraEntity): String = {
+    AgoraConfig.methodsUrl + entity.namespace.get + "/" + entity.name.get + "/" + entity.snapshotId.get
+  }
+
+  def addUrl(entity: AgoraEntity): AgoraEntity = {
+    entity.copy(url = Option(agoraUrl(entity)))
+  }
+
+  def insert(agoraEntity: AgoraEntity): AgoraEntity = {
+    AgoraDao.createAgoraDao.insert(agoraEntity).copy(url = Option(agoraUrl(agoraEntity)))
+  }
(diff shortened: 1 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import org.broadinstitute.dsde.agora.server.AgoraConfig
+import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+/**
+ * Created by dshiga on 5/13/15.
+ */
+object AgoraBusiness extends AgoraDao {
+
```
jacarey: We shouldn't need to extend AgoraDao here.. you should be able to import AgoraDao._ and have access to the factory methods.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -0,0 +1,36 @@
...
+  def agoraUrl(entity: AgoraEntity): String = {
+    AgoraConfig.methodsUrl + entity.namespace.get + "/" + entity.name.get + "/" + entity.snapshotId.get
+  }
+
+  def addUrl(entity: AgoraEntity): AgoraEntity = {
+    entity.copy(url = Option(agoraUrl(entity)))
+  }
+
+  override def insert(agoraEntity: AgoraEntity): AgoraEntity = {
+    AgoraDao.createAgoraDao.insert(agoraEntity).copy(url = Option(agoraUrl(agoraEntity)))
```
jacarey: What do we gain from url being immutable and being forced to copy vs url being a var and just setting it?

dshiga: Perhaps nothing. I was trying to do the Scala thing where you try to keep everything immutable, but maybe it's overkill in this case.

jacarey: I'm not against it. As they would say "Mutability is an optimization". I'm fine leaving it as is but we should note that this might be a target for optimization should performance become an issue.

jacarey: I would also note that we should probably be consistent. So if we want to use this pattern until a later time we should use it for snapshot id and owner as well. I will make those changes to my pull request.

PR #37, opened 2015-05-14
Fix url building when data is missing, fix createDate

dshiga:
Fix url building when namespace, name, or id is missing (return empty string). Populate createDate when inserting a new method.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -7,7 +7,14 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 object AgoraBusiness {
 
   def agoraUrl(entity: AgoraEntity): String = {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -10,7 +10,7 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 object AgoraBusiness {
 
   def agoraUrl(entity: AgoraEntity): String = {
-    AgoraConfig.methodsUrl + entity.namespace.get + "/" + entity.name.get + "/" + entity.snapshotId.get
+    AgoraConfig.methodsUrl + entity.namespace.getOrElse("") + "/" + entity.name.getOrElse("") + "/" + entity.snapshotId.getOrElse("")
```
jacarey: Should we ever even get here if any of these are None?  Wouldn't we be returning a nonsense URL if that was the case?

jacarey:
+1

PR #38, opened 2015-05-15
Return 4xx errors in JSON format to fix swagger UI

dshiga:
We were returning 4xx error responses in plain text format, which broke the swagger UI and hid the actual error message.

(Details: The swagger UI sends an "accepts" header requiring responses to be in JSON. That's ok for success responses, which are in JSON. But for 404 errors, for example, we were attempting to return a plain text error message. Spray tries to be smart about this, so instead of returning the plain text 404 error message, it returns a 406 to indicate that the response format does not match the "accepts" header.)

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraError.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraError.scala
@@ -1,0 +1,13 @@
+package org.broadinstitute.dsde.agora.server.model
+
+import com.wordnik.swagger.annotations.{ApiModelProperty, ApiModel}
+
+import scala.annotation.meta.field
+
+/**
+ * Created by dshiga on 5/15/15.
+ */
+@ApiModel(value = "Agora Method")
+case class AgoraError(@(ApiModelProperty@field)(required = true, value = "The error message")
+                       error: String
+                        )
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraError.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import com.wordnik.swagger.annotations.{ApiModelProperty, ApiModel}
+
+import scala.annotation.meta.field
+
+/**
+ * Created by dshiga on 5/15/15.
+ */
+@ApiModel(value = "Agora Method")
+case class AgoraError(@(ApiModelProperty@field)(required = true, value = "The error message")
```
bradtaylor: Do we want this to be a case class with a string message, or do we want it to be an agora-specific exception, and just have that serialized to JSON? Thought I'd seen examples of spray json serializing exceptions

dshiga:
I'm not sure what happened to Brad's comment on the AgoraError case class, which was:
"Do we want this to be a case class with a string message, or do we want it to be an agora-specific exception, and just have that serialized to JSON? Thought I'd seen examples of spray json serializing exceptions"

That's a good point and I'll look into serializing exceptions, but am going to merge this now to relieve the problem in the short term.

jacarey:
I think having a generalized exception that can be serialized to json is probably a good idea. It allows us to bubble up errors and let the layer above decide what to do with it.

PR #39, opened 2015-05-18
Validate metadata before inserting a method

dshiga:


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsService.scala
```diff
@@ -90,9 +91,12 @@ trait MethodsService extends HttpService with PerRequestCreator with AgoraDirect
       post {
         commonNameFromCookie() { commonName =>
           entity(as[AgoraEntity]) { agoraEntity =>
-            requestContext =>
-              val entityWithOwner = agoraEntity.copy(owner = Option(commonName))
-              perRequest(requestContext, methodsAddHandlerProps, ServiceMessages.Add(requestContext, entityWithOwner))
+            val validation = AgoraValidation.validateMetadata(agoraEntity)
+            validate(validation.valid, validation.message) {
```
bradtaylor: Not knowing much about this directive, how is the route rejection with ValidationRejection handled? Does the request ultimately get completed with 400?

dshiga: Yes, it returns a 400 with a plain text response containing the error message.

bradtaylor: I think we can supply a rejection handler wherein we can wrap the message in AgoraError in order to json serialize. Should be in spray routing docs

dshiga:
I looked into spray Rejections per your suggestion, and ended up making a custom Rejection class called AgoraValidationRejection. It contains an AgoraValidation object, which is basically just a Seq[String] of error messages.

I also set things up so that the AgoraValidationRejection gets serialized properly as JSON back to the user. They get a 400 back with JSON that looks like this example: { "error": ["Namespace is required", "Name is required"]}.

(The standard ValidationRejection provided by spray only lets you specify a single String as a message, but I think it will be important to support returning multiple validation messages back to the user at once).

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/validation/AgoraValidation.scala
```diff
@@ -0,0 +1,35 @@
...
+    }
+    validation = entity.synopsis.forall(_.length() <= 80) match {
+      case false => validation.addError("Synopsis must be 80 characters or less")
+      case true => validation
+    }
+    validation
+  }
+}
+
+case class AgoraValidation(messages: Seq[String] = Seq[String]()) {
```
jacarey: Use Seq.empty[String] instead

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/validation/AgoraValidation.scala
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Created by dshiga on 5/18/15.
+ */
+
+object AgoraValidation {
+  def validateMetadata(entity: AgoraEntity): AgoraValidation = {
+    var validation = AgoraValidation()
+    validation = entity.namespace.exists(_.trim.nonEmpty) match {
+      case false => validation.addError("Namespace is required")
+      case true => validation
```
jacarey: since validation is var anyways it might be a bit cleaner to just use an if here and add the message.

if(!entity.namespace.exists(_.trim.nonEmpty)) validation.addError("Namespace is required)

jacarey:
:+1:  Just needs a squash. Looks good!

PR #42, opened 2015-05-19
Added the ability to specify included and excluded fields for db proj…

jacarey:
…ection.

Validate we aren't mixing include and exclude fields.
Validate we aren't excluding required fields.
Make sure we add required fields to user specified includes.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
@@ -11,7 +11,8 @@ import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
 import org.broadinstitute.dsde.agora.server.dataaccess.mongo.AgoraMongoClient._
 import org.broadinstitute.dsde.agora.server.dataaccess.mongo.AgoraMongoDao._
 import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
-import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+import org.broadinstitute.dsde.agora.server.model.AgoraProjectionDefaults._
+import org.broadinstitute.dsde.agora.server.model.{AgoraEntity, AgoraEntityProjection}
 import spray.json._
 
 object AgoraMongoDao {
@@ -19,6 +20,7 @@ object AgoraMongoDao {
   val CounterCollectionName = "counters"
   val CounterSequenceField = "seq"
   val KeySeparator = ":"
+  val DefaultFindProjection = Option(new AgoraEntityProjection(Seq.empty[String], Seq[String]("payload", "documentation")))
 
   def EntityToMongoDbObject(entity: AgoraEntity): DBObject = JSON.parse(entity.toJson.toString()).asInstanceOf[DBObject]
 
@@ -53,7 +55,7 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
     val counterQuery = MongoDbIdField $eq counterId
 
     //if we don't have a sequence create one
-    if (counterCollection.findOne(counterQuery) == None) {
+    if (counterCollection.findOne(counterQuery).isEmpty) {
       counterCollection.insert(MongoDBObject(MongoDbIdField -> counterId, CounterSequenceField -> 0))
     }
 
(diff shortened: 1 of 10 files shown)


jacarey:
Some members got rearranged. I think checked "Rearrage members" by accident when I commited.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -25,4 +26,15 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = false, value = "The
                        payload: Option[String] = None,
                        @(ApiModelProperty@field)(required = false, value = "URI for method details")
                        url: Option[String] = None
-                        )
\ No newline at end of file
+                        )
+
+object AgoraProjectionDefaults {
+  val RequiredProjectionFields = Seq[String]("namespace", "name", "id")
```
gbggrant: Should this be snapshotId?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -25,4 +26,15 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = false, value = "The
...
+}
+
+case class AgoraEntityProjection(includedFields: Seq[String], excludedFields: Seq[String]) {
+  require(excludedFields.intersect(RequiredProjectionFields).isEmpty)
+  require(includedFields.isEmpty || excludedFields.isEmpty)
+
```
gbggrant: Should it be an error condition if a field is both included and excluded?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -83,6 +76,33 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
     }
   }
 
+  override def find(entity: AgoraEntity, projection: Option[AgoraEntityProjection]): Seq[AgoraEntity] = {
+    projection match {
+      case Some(project) => find(EntityToMongoDbObject(entity), projection)
```
gbggrant: Should use project and not projection (doesn't really make a diff, I think - but that's what you do elsewhere).

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: indeed it should

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: I think the condition of "one list has to be empty" is more strict so the above condition would never fall through.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
jacarey: mostly did this to avoid shadowing.. but I agree a different name seems more appropriate

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -193,11 +201,30 @@ with AgoraTestData with BeforeAndAfterAll {
...
+      + "&name=" + name2.get
+      + "&excludedField=synopsis&excludedField=documentation&excludedField=createDate&excludedField=payload") ~>
+      methodsService.queryRoute ~> check {
+      handleError(
+        entity.as[Seq[AgoraEntity]],
+        (entities: Seq[AgoraEntity]) =>
+          assert(entities === excludeProjection(Seq(testEntity3WithId, testEntity4WithId, testEntity5WithId, testEntity6WithId, testEntity7WithId)))
+      )
+      assert(status === OK)
     }
```
gbggrant: Probably should add test to check if excludedFields includes any of the required project fields.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
gbggrant: Yeah - sorry was reading that the wrong way.  Forget the second line I added to the note.

gbggrant:
Looks good.  :100: / :100:

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
jacarey: We should be catching this ... but I can add a test that verifies the case class validation fails.

PR #45, opened 2015-05-20
Add agora.conf to the classpath to get openam working.

jacarey:


bradtaylor:
Needs a comment that this is an interim solution.
Otherwise, thumbsup

PR #46, opened 2015-05-20
Added the ability to get user name instead of common name.

jacarey:
It would be nice to push this down to vault common.


gbggrant:
Looks good. :+1:

PR #48, opened 2015-05-21
We are going to copy agora.conf from /etc/agora.conf in the jenkins b…

jacarey:
…uild to application.conf such that it gets built in to the jar.

Also fix the tests that were relying on entity ordering.


gbggrant:
bueno!  :+1:

PR #52, opened 2015-05-21
Validate include and exclude fields up in MethodsService

dshiga:
So we can return a 400 instead of a 500 if invalid.


jacarey:
Excellent! Ship it!

PR #56, opened 2015-05-22
Get the email from openam since the username is currently set to the …

jacarey:
…uuid

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraOpenAMDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraOpenAMDirectives.scala
@@ -1,6 +1,7 @@
 
 package org.broadinstitute.dsde.agora.server.webservice
 
+import org.broadinstitute.dsde.agora.server.webservice.util.AgoraOpenAMClient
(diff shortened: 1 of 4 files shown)


jacarey:
AgoraOpenAMClient.scala has methods that should make their way down in to vault-common... but for now they exist in Agora.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraOpenAMDirectives.scala
```diff
@@ -27,8 +28,9 @@ trait AgoraOpenAMDirectives extends AgoraDirectives {
   def usernameFromToken(token: String)(implicit ec: ExecutionContext): Directive1[String] = {
     val userNameFuture = for {
       id <- OpenAMClient.lookupIdFromSession(OpenAMConfig.deploymentUri, token)
-      usernameCN <- OpenAMClient.lookupUsernameCN(OpenAMConfig.deploymentUri, token, id.id, id.realm)
-    } yield usernameCN.username
+      usernameCN <- AgoraOpenAMClient.lookupUserInfo(OpenAMConfig.deploymentUri, token, id.id, id.realm)
```
dshiga: Should this be renamed userInfo?

PR #59, opened 2015-05-28
Added slf4j adapter to log akka messages.

jacarey:
Also I rearranged the dependencies so they are alphabetical... as the list grows it makes it easier find what you are looking for.


bradtaylor:
image

bradtaylor:
Do it

PR #61, opened 2015-06-08
Added task config endpoints.

jacarey:
Added entity typing and the ability to search across collections.
Refactor tests in to snack sized bites.


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -32,8 +32,14 @@ class MethodsAddHandler extends Actor {
...
   private def validatePayload(agoraEntity: AgoraEntity): Unit = {
-    WdlBinding.getAst(agoraEntity.payload.get, agoraEntity.name.get)
-//    WdlBinding.process(agoraEntity.payload.get, agoraEntity.name.get, AgoraBusiness.importResolver)
+    agoraEntity.entityType.get match {
+      case AgoraEntityType.Task =>
+        WdlBinding.getAst(agoraEntity.payload.get, agoraEntity.name.get)
+      case AgoraEntityType.Workflow =>
+        WdlBinding.getAst(agoraEntity.payload.get, agoraEntity.name.get)
+      case AgoraEntityType.Configuration =>
+      //add config validation here
```
bradtaylor: syntactically this is JSON validation

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -27,7 +27,17 @@ object AgoraMongoDao {
...
-class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
+/**
+ * The data access object for agora's mongo database. If more than one collection is specified we will query all
+ * collections. However, only a single collection is allowed for doing insert operations.
+ * @param collections The collections to query. In order to insert a single colleciton must be specified.
+ */
+class AgoraMongoDao(collections: Seq[MongoCollection]) extends AgoraDao {
+
+  def getSingleCollection: MongoCollection = {
+    if (collections.size != 1) throw new IllegalArgumentException("Inserts can only be run on a single collection")
```
gbggrant: Kind of a deceptive exception message - the method is getSingleCollection and could be used elsewhere than insert.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
jacarey: yup... but might include reference validation as well.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
jacarey: agreed

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
bradtaylor: Probably a separate story

gbggrant:
Looks fine by me, but think that @bradtaylor is still looking at it.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.broadinstitute.dsde.agora.server.model.{AgoraEntityType, AgoraEntity, AgoraEntityProjection}
+import org.broadinstitute.dsde.agora.server.webservice.util.ServiceMessages
+import org.broadinstitute.dsde.agora.server.webservice.validation.{AgoraValidation, AgoraValidationRejection}
+import org.joda.time.DateTime
+import spray.routing.{HttpService, Route}
+
+trait AgoraService extends HttpService with PerRequestCreator with AgoraDirectives {
+  this: ServiceHandlerProps =>
+
+  private implicit val ec = actorRefFactory.dispatcher
```
bradtaylor: What is this?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -0,0 +1,75 @@
...
+        requestContext =>
+          perRequest(requestContext, queryHandlerProps, ServiceMessages.QueryByNamespaceNameSnapshotId(requestContext, namespace, name, snapshotId.toInt,
+          AgoraEntityType.byPath(path)))
+      }
+    }
+
+  def queryRoute =
+    path(path) {
+      get {
+        parameters("namespace".?, "name".?, "snapshotId".as[Int].?,
```
bradtaylor: Can we pretty-print this a little more cleanly? The 'as' on EntityType and on the 'as' on the whole parameters object are kinda close together

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
jacarey: This is the execution context for the magnet pattern the vault uses for vault commons. It is needed for the agora directives to get username from the token. I can rename it to executionContext

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
jacarey: Am I pretty enough for you now Brad?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -48,7 +48,10 @@ object AgoraBusiness {
   def importResolver(importString: String) : String = {
     val importStringPattern = """^methods\:\/\/(\S+)\.(\S+).(\d+)""".r
     importString match {
-      case importStringPattern(namespace, name, snapshotId) => AgoraBusiness.findSingle(namespace, name, snapshotId.toInt).getOrElse(new AgoraEntity()).payload.getOrElse("")
+      case importStringPattern(namespace, name, snapshotId) =>
+        //hard coded for tasks...we need to ask wdl parser which type we have
```
bradtaylor: What is going on here? Tasks cannot import other tasks/workflows. But workflows CAN import other workflows. It seems like hardcoding the new AgoraEntity() as a task is fine because we are just using it as something with an empty payload in the subsequent chained calls.

jacarey: Ah I think this comment is out of date. I managed to work around the issue by allow a user to specify a Seq of types to search. Previously it had to specify the type in order to search and so we would have had to have know that ahead of time.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraDao.scala
```diff
@@ -4,16 +4,33 @@ package org.broadinstitute.dsde.agora.server.dataaccess
...
   def createAgoraDao(collection: MongoCollection): AgoraDao = {
-    new AgoraMongoDao(collection)
+    new AgoraMongoDao(Seq(collection))
+  }
+
+  def createAgoraDao(entityTypes: Seq[AgoraEntityType.EntityType]): AgoraDao ={
+    new AgoraMongoDao((for(entityType <- entityTypes) yield getCollectionsByEntityType(Option(entityType))).flatten)
+  }
+
+  def getCollectionsByEntityType(entityType: Option[AgoraEntityType.EntityType]) : Seq[MongoCollection] ={
```
bradtaylor: Does this belong here, or in the client with the other MongoCollection returning functions?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraDao.scala
```diff
@@ -4,16 +4,33 @@ package org.broadinstitute.dsde.agora.server.dataaccess
...
+    new AgoraMongoDao(getCollectionsByEntityType(entityType))
   }
 
   def createAgoraDao(collection: MongoCollection): AgoraDao = {
-    new AgoraMongoDao(collection)
+    new AgoraMongoDao(Seq(collection))
+  }
+
+  def createAgoraDao(entityTypes: Seq[AgoraEntityType.EntityType]): AgoraDao ={
+    new AgoraMongoDao((for(entityType <- entityTypes) yield getCollectionsByEntityType(Option(entityType))).flatten)
```
bradtaylor: Can/should you use entityTypes.map instead of for-yield?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraDao.scala
jacarey: Good point.. I'll move it.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -84,7 +96,9 @@ class AgoraMongoDao(collection: MongoCollection) extends AgoraDao {
   }
 
   def find(query: Imports.DBObject, projection: Option[AgoraEntityProjection]) = {
-    (for (dbObject <- collection.find(query, projectionToDBProjections(projection))) yield MongoDbObjectToEntity(dbObject)).toVector
+    (for (collection <- collections) yield {
```
bradtaylor: Similar question re. for-yield vs map

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraDao.scala
jacarey: yeah I think .flatMap would be cleaner.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -25,7 +38,8 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = false, value = "The
                        @(ApiModelProperty@field)(required = false, value = "The method payload")
                        payload: Option[String] = None,
                        @(ApiModelProperty@field)(required = false, value = "URI for method details")
-                       url: Option[String] = None
+                       url: Option[String] = None,
+                       entityType: Option[AgoraEntityType.EntityType]= None
```
bradtaylor: We are packing entityType into metadata? I thought it was just being encoded by storing it in different underlying collections and being asked-for by different endpoints. If we do want it in the metadata packet for an Entity, should it have a swagger ApiModelProperty?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  def path: String
+
+  def routes = queryByNamespaceNameSnapshotIdRoute ~ queryRoute ~ postRoute
+
+  def queryByNamespaceNameSnapshotIdRoute =
+    path(path / Segment / Segment / Segment) { (namespace, name, snapshotId) =>
+      get {
+        requestContext =>
+          perRequest(requestContext, queryHandlerProps, ServiceMessages.QueryByNamespaceNameSnapshotId(requestContext, namespace, name, snapshotId.toInt,
```
bradtaylor: pretty print the perRequest() ServiceMessages...() calls please?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -0,0 +1,85 @@
...
+          "namespace".?,
+          "name".?,
+          "snapshotId".as[Int].?,
+          "synopsis".?,
+          "documentation".?,
+          "owner".?,
+          "createDate".as[DateTime].?,
+          "payload".?,
+          "url".?,
+          "entityType".as[AgoraEntityType.EntityType].?
```
bradtaylor: Why can we search by entity type in metadata?  Wouldn't we be calling GET with parameters on different endpoints?  What would it mean if I searched /configurations with entityType=methods?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -0,0 +1,85 @@
...
+          }
+        }
+      }
+    }
+
+  def postRoute =
+    path(path) {
+      post {
+        usernameFromCookie() { commonName =>
+          entity(as[AgoraEntity]) { agoraEntity =>
```
bradtaylor: what happens if I post an entity with "entityType"="configuration" to /methods?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: This is more for when an entity moves through persistence -> domain -> web service. We need each layer to know what type it is without having to infer it.

I suppose it doesn't need to be stored in the database directly as it can be inferred from the collection, but it would be more work to exclude it. 

For the methods post end point we would need to add it as an api model property otherwise we would need to pull it out of the WDL. I'm fine with either... opinions?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/configurations/ConfigurationsService.scala
```diff
@@ -0,0 +1,71 @@
...
+    new ApiImplicitParam(name = "namespace", required = true, dataType = "string", paramType = "path", value = "Namespace"),
+    new ApiImplicitParam(name = "name", required = true, dataType = "string", paramType = "path", value = "Name"),
+    new ApiImplicitParam(name = "snapshotId", required = true, dataType = "string", paramType = "path", value = "SnapshotId")
+  ))
+  @ApiResponses(Array(
+    new ApiResponse(code = 200, message = "Successful Request"),
+    new ApiResponse(code = 404, message = "Task Configuration Not Found"),
+    new ApiResponse(code = 500, message = "Internal Error")
+  ))
+  override def queryByNamespaceNameSnapshotIdRoute = super.queryByNamespaceNameSnapshotIdRoute
```
bradtaylor: Why override a def and then just call super?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
jacarey: Technically that is valid... though it would never return any information. However unless we want to do parameter map parsing we have to be be able to parse every possible search field in order to use case class extrapolation :(

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
jacarey: See my other comment about searching entity type on the wrong end point.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/configurations/ConfigurationsService.scala
```diff
@@ -0,0 +1,71 @@
...
+import com.wordnik.swagger.annotations._
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+import org.broadinstitute.dsde.agora.server.webservice.{ServiceHandlerProps, AgoraService}
+import org.broadinstitute.dsde.agora.server.webservice.util.ApiUtil
+
+trait ConfigurationsService extends AgoraService {
+  this: ServiceHandlerProps =>
+  // Require a concrete ServiceHandlerProps creator to be mixed in
+
+  def path = ApiUtil.Configurations.path
```
bradtaylor: Should be override def?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/methods/MethodsService.scala
```diff
@@ -1,26 +1,18 @@
...
 
 @Api(value = "/methods", description = "Method Service", produces = "application/json", position = 1)
-trait MethodsService extends HttpService with PerRequestCreator with AgoraDirectives {
+trait MethodsService extends AgoraService {
   this: ServiceHandlerProps =>
   // Require a concrete ServiceHandlerProps creator to be mixed in
 
-  import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
-  import spray.httpx.SprayJsonSupport._
+  def path = ApiUtil.Methods.path
```
bradtaylor: override?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/configurations/ConfigurationsService.scala
jacarey: exactly... not ideal but swagger doesn't give us a choice. If we could have non static annotation strings we wouldn't need to do this.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ServiceHandlerProps.scala
```diff
@@ -0,0 +1,10 @@
+
+package org.broadinstitute.dsde.agora.server.webservice
+
+import akka.actor.Props
+
+trait ServiceHandlerProps {
```
bradtaylor: Since we're not overriding with MethodsServiceHandlerProps and ConfigsServiceHandlerProps, and the underlying handlers have been made generic, does this need to be it's own class?  Or can it become part of AgoraService?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/ServiceMessages.scala
```diff
@@ -8,7 +8,11 @@ import spray.routing.RequestContext
...
 object ServiceMessages {
 
-  case class QueryByNamespaceNameSnapshotId(requestContext: RequestContext, namespace: String, name: String, snapshotId: Int)
+  case class QueryByNamespaceNameSnapshotId(requestContext: RequestContext,
+                                            namespace: String,
+                                            name: String,
+                                            snapshotId: Int,
+                                            entityType: Seq[AgoraEntityType.EntityType])
 
   case class Query(requestContext: RequestContext, agoraSearch: AgoraEntity, agoraProjection: Option[AgoraEntityProjection])
```
bradtaylor: Pretty print this, and add, since we did the first one?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ServiceHandlerProps.scala
jacarey: Good question. It was separate before so I'm not sure what the initial reason was. (I think it was a pattern we had used before in Apollo)

bradtaylor: Initial reason was so it could be overriden. In apollo, by test-vs-prod Props. Here, we were thinking different endpoint props. I'd say fold it in to AgoraService

PR #64, opened 2015-06-12
Added basic agora acls as a bit mask.

jacarey:


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraAcl.scala
```diff
@@ -0,0 +1,26 @@
...
+  def canRead: Boolean = (permissions & Read) != 0
+
+  def canWrite: Boolean = (permissions & Write) != 0
+
+  def canCreate: Boolean = (permissions & Create) != 0
+
+  def canRedact: Boolean = (permissions & Redact) != 0
+
+  def canManage: Boolean = (permissions & Manage) != 0
+}
```
dshiga: This is neat, but I worry that it's less straightforward than using a Set, especially for creating and modifying acls. I would favor using Sets unless we have a good reason to prefer this way.

jacarey: So you don't feel that AgoraAcl(Read) or AgoraAcl(Read|Write) are explicit enough and would be in favor of 
new AgoraAcl
AgoraAcl.setRead(true)
AgoraAcl.setWrite(true)

?

dshiga: I was wondering more about stuff like:
1) create AgoraAcl with Read, Write, Create, and Redact
2) given an AgoraAcl, make a new one minus Create and Redact

If we used sets, 1) would be something like acl = Set(Read, Write, Create, Redact), but with the bit field would be acl = AgoraAcl(Read | Write | Create | Redact).

With sets, 2) would be acl = acl - Set(Create, Redact), but with a bit field would be acl = AgoraAcl(acl.permissions & (31 ^ Create) & (31 ^ Redact)) ... I think.

Isn't it easier to work with sets instead of a bit field, especially since we don't need to store the AgoraAcl?

jacarey: Ah ok I misunderstood what you were saying. It is fair point that if we aren't storing things we don't need to worry about the efficiency of a bit field. When I think permissions I think bitmask so that was my initial thought for implementation since it is probably the most common and mimics posix permissions.

Personally I don't find acl = AgoraAcl(Read | Write | Create |Redact) to be anymore onerous than Set(Read, Write, Create, Redact) in fact you could create a var arg constructor that can do just that AgoraAcl(Read, Write, Create, Redact)

For 2 I'm not sure it is much more complicated either just and the not value
so AgoraAcl(acl.permisions & ~Create & ~Redact).

If we ever did need to store them serialization becomes easier and you also get more type safety using a case class instead of a generic Set of ints.

Does anyone else have an opinion?

gbggrant: I like the simplicity of the bit masks.

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraAccessControlTest.scala
```diff
@@ -0,0 +1,61 @@
...
+  "Agora" should "return true for all permissions if they have full access" in {
+    val acl = AgoraAcl(All)
+    assert(acl.canRead === true)
+    assert(acl.canWrite === true)
+    assert(acl.canCreate === true)
+    assert(acl.canRedact === true)
+    assert(acl.canManage === true)
+  }
+
+  "Agora" should "be able to construct and ACL using var arg permissions" in {
```
bradtaylor: construct an, not and

PR #67, opened 2015-06-19
Business layer authorization.

jacarey:
I gave up on fancy DI.
Added basic read authorization test.


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -27,21 +26,38 @@ object AgoraBusiness {
...
+    val entities = AgoraDao.createAgoraDao(entityTypes).find(agoraSearch, agoraProjection).map(entity => entity.copy(url = Option(agoraUrl(entity))))
+    filterByReadPermissions(authorizationProvider.authorizationsForEntities(entities))
   }
 
   def findSingle(namespace: String,
                  name: String,
                  snapshotId: Int,
                  entityTypes: Seq[AgoraEntityType.EntityType]): Option[AgoraEntity] = {
-    AgoraDao.createAgoraDao(entityTypes).findSingle(namespace, name, snapshotId).map(entity => entity.copy(url = Option(agoraUrl(entity))))
+    val optEntity = AgoraDao.createAgoraDao(entityTypes).findSingle(namespace, name, snapshotId).map(entity => entity.copy(url = Option(agoraUrl(entity))))
```
bradtaylor: Should we be doing the None check before doing the entity.copy and setting the agoraUrl? Seems like we can avoid doing this if the entity is None.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -22,7 +22,10 @@ object AgoraMongoDao {
   val KeySeparator = ":"
   val DefaultFindProjection = Option(new AgoraEntityProjection(Seq.empty[String], Seq[String]("payload", "documentation")))
 
-  def EntityToMongoDbObject(entity: AgoraEntity): DBObject = JSON.parse(entity.toJson.toString()).asInstanceOf[DBObject]
+  def EntityToMongoDbObject(entity: AgoraEntity): DBObject = {
+    //we need to remove url if it exists since we don't store it in the document
```
bradtaylor: Why don't we?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -80,7 +83,7 @@ class AgoraMongoDao(collections: Seq[MongoCollection]) extends AgoraDao {
   }
 
   override def findSingle(entity: AgoraEntity): Option[AgoraEntity] = {
-    val entityVector = find(EntityToMongoDbObject(entity), None)
+    val entityVector = find(entity, None)
```
bradtaylor: Why don't we need to find a MongoDbObject?  What changed?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -41,7 +41,9 @@ case class AgoraEntity(@(ApiModelProperty@field)(required = false, value = "The
                        url: Option[String] = None,
                        @(ApiModelProperty@field)(dataType = "string", required = false, value = "The type of the entity (Task, Workflow, Configuration)", allowableValues = "Task,Workflow,Configuration")
                        entityType: Option[AgoraEntityType.EntityType] = None
-                        )
+                        ) {
+  def getUniqueIdentifier: String = namespace + ":" + name + ":" + snapshotId
```
bradtaylor: Do we want this as a method on the case class or do we want to put it as a static in the companion object?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/gcs/GcsAuthorizationProvider.scala
```diff
@@ -0,0 +1,19 @@
+
+package org.broadinstitute.dsde.agora.server.dataaccess.gcs
+
+import org.broadinstitute.dsde.agora.server.business.AgoraPermissions._
+import org.broadinstitute.dsde.agora.server.business.{AuthorizationProvider, AgoraAcl, AuthorizedAgoraEntity}
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+object GcsAuthorizationProvider extends AuthorizationProvider {
+  override def authorizationsForEntity(agoraEntity: AgoraEntity): AuthorizedAgoraEntity = {
+    //TODO call google here and do GACL -> AgAcl translation
```
bradtaylor: You are going to need to pass to Google the common-name of the user whose permissions you want to check. This is not the owner of the method, but the user making the request. So we will need to add the usernameFromCookie() directive at AgoraService:95 to the queryRoute and queryByNamespaceNameSnapshotIdRoute, and pass this down into here.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
```diff
@@ -4,30 +4,33 @@ import akka.actor.Props
...
 
 object ApiServiceActor {
-  def props: Props = Props(classOf[ApiServiceActor])
+  def props(authorizationProvider: AuthorizationProvider): Props = Props(classOf[ApiServiceActor], authorizationProvider)
 }
 
-class ApiServiceActor extends HttpServiceActor {
+class ApiServiceActor(authorization: AuthorizationProvider) extends HttpServiceActor {
+
+  import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
```
bradtaylor: Why move these imports inside the class? In general, when do you want to do this in scala/akka?

geoffjentry: Generally you'd do if you want to ensure particular implicits are only in scope in specific places

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/authorization/TestAuthorizationProvider.scala
```diff
@@ -0,0 +1,24 @@
...
+
+//Test authorization implementation. Add authorizations directly to the hash map for lookup. If the entity is not found it is assumed you
+//have no permissions on that entity.
+object TestAuthorizationProvider extends AuthorizationProvider {
+  val localPermissions: mutable.Map[String, AgoraAcl] = new mutable.HashMap[String, AgoraAcl]()
+
+  def addLocalPermissions(entityId: String, authorization: AgoraAcl) = localPermissions.put(entityId, authorization)
+
+  override def authorizationsForEntity(agoraEntity: AgoraEntity): AuthorizedAgoraEntity = {
+    AuthorizedAgoraEntity(agoraEntity, localPermissions.getOrElse(agoraEntity.getUniqueIdentifier, AgoraAcl(All)))
```
bradtaylor: Seems like if entity is not found in localPermissions, it gets All permission. But the comment above makes it sound like it should get No permission?

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -18,9 +18,7 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
         assert(entity.namespace === namespace1)
         assert(entity.name === name1)
         assert(entity.synopsis === synopsis3)
-        assert(entity.documentation === documentation1)
```
bradtaylor: Why aren't we checking docs and payload anymore?

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
```diff
@@ -107,11 +105,16 @@ class AgoraMethodsSpec extends ApiServiceSpec {
   }
 
   "Agora" should "store 10kb of github markdown as method documentation and return it without alteration" in {
-    Post(ApiUtil.Methods.withLeadingSlash, testAgoraEntityBigDoc) ~>
+    Post(ApiUtil.Methods.withLeadingSlash + "?includedField=documentation", testAgoraEntityBigDoc) ~>
```
bradtaylor: You can pass the includedField parameter to our Post route?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
jacarey: We could though it clutters things a little since you would be doing the match here as well as below.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/gcs/GcsAuthorizationProvider.scala
jacarey: Good catch. We can do that when we wire things up.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
```diff
@@ -107,11 +105,16 @@ class AgoraMethodsSpec extends ApiServiceSpec {
...
       methodsService.postRoute ~> check {
       handleError(
         entity.as[AgoraEntity],
-        (entity: AgoraEntity) => assert(entity.documentation.get === bigDocumentation.get)
+        (entity: AgoraEntity) => {
+          val searchEntity = new AgoraEntity(name = entity.name, namespace = entity.namespace, snapshotId = entity.snapshotId)
```
bradtaylor: Why do we need to perform a separate search? Shouldn't the return packet of the method post desearialize to an entity that includes the documentation?

I'm a little unclear what's going on with the changes in this test

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
jacarey: I think the reasoning is that it has configurable components. (port and hostname for instance) Because of that we would potentially be storing urls that become invalid and we have no way of changing them for older entities since we are write once.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
jacarey: findSingle should be calling the find(entity, projection) method so that it can get the default projections instead of returning the whole object. Alternately we could have it return the fully projected entity, but then we would have two different behaviours for the two find methods.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
jacarey: Currently I'm only using this in testing frameworks so it might make sense to just push it down to a testing companion object.

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/authorization/TestAuthorizationProvider.scala
jacarey: Yeah I changed this. I will update the comment.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
jacarey: nope was testing. need to remove this.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
bradtaylor: Good reason!

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
jacarey: This change is related to the default projection on findSingle. Are you of the opinion that findSingle should return the entire entity or just the projected entity like find does? In either case I think they should be consistent with each other.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
jacarey: also related to findSingle changes.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
bradtaylor: I think if you are hitting GET namespace/name/id you should get the whole entity, docs and payload included. If you are referencing a specific method, don't you want the entire method? Otherwise the only way for a caller to look at their documentation is to specify:
GET methods?namespace=mine&name=myMethod&snapshotId=X&includedField=documentation
vs
GET methods/mine/myMethod/snapshotId

I thought that was part of the motivation for keeping 2 GET endpoints- one for a list of truncated entities, configurable with entity projection. The other for getting the full details of a single method.

or does changing findSingle not affect that behavior?  I'm less concerned about the guts than about the user-facing endpoints

jacarey: Ah good points.
The issue is that findSingle is used both on methods?namespace?name?id and on the post so will get one behavior or the other for both methods since the converge here.

I'm going to revert changes to findSingle since I think that use case of getting the entire entity is more important.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
bradtaylor: The following seems like it would save you the map, and just move the entityCopy within the check. Seems like we're double-checking for None by doing both. I was hoping it would be /less/ cluttered. But looking at it, it's probably 6 of one...
```
    val optEntity = AgoraDao.createAgoraDao(entityTypes).findSingle(namespace, name, snapshotId)
    optEntity match {
      case Some(foundEntity) => filterByReadPermissions(authorizationProvider.authorizationsForEntity(foundEntity.copy(url = Option(agoraUrl(foundEntity)))))
      case None => None
    }
  }
```

jacarey: Ah... now I can smell the cake you're baking. I like it.

PR #68, opened 2015-06-19
Create translator between Google ACLs and Agora permissions

dshiga:
We have two levels of permissions: those on a namespace, and those on a method.

We populate namespace permissions solely by looking at the corresponding gcs bucket role.
We populate method permissions solely by looking at the corresponding gcs object role.

There are 5 possible permissions on both namespaces and methods (Read, Write, Create, Redact, Manage). All 5 can be populated on namespaces. For methods, only Read and Manage will ever be populated as long as we're using gcs acls as our back end (because there is no WRITER permission on object acls).

The translator does not try to model the idea that namespace Read implies method Read. It's up to callers to know that a user should be allowed to read a method if they have Read on the namespace but not on the method. Similarly, Write will never be populated on a method, and it's up to callers to check for namespace level Write to decide if the user can modify a method.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/acls/GcsRole.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/acls/GcsRole.scala
@@ -1,0 +1,26 @@
+package org.broadinstitute.dsde.agora.server.acls
+
+import org.broadinstitute.dsde.agora.server.acls.GcsRole._
+
+object GcsRole {
+  val Reader = "READER"
+  val Writer = "WRITER"
+  val Owner = "OWNER"
+  val Nothing = "NOTHING"
+}
+
+case class GcsBucketRole(role: String) {
+
+  def isReader: Boolean = role.equals(Reader) || role.equals(Writer) || role.equals(Owner)
+
+  def isWriter: Boolean = role.equals(Writer) || role.equals(Owner)
+
+  def isOwner: Boolean = role.equals(Owner)
+}
+
+case class GcsObjectRole(role: String) {
+
+  def isReader: Boolean = role.equals(Reader) || role.equals(Owner)
+
+  def isOwner: Boolean = role.equals(Owner)
+}
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
@@ -1,0 +1,42 @@
+package org.broadinstitute.dsde.agora.server.acls
+
+import org.broadinstitute.dsde.agora.server.business.AgoraPermissions
+import org.broadinstitute.dsde.agora.server.business.AgoraPermissions._
+import org.broadinstitute.dsde.agora.server.acls.GcsRole._
+
+object RoleTranslator {
+
+  def gcsBucketToNamespacePermissions(bucketRole: GcsBucketRole): AgoraPermissions = {
+    bucketRole match {
+      case GcsBucketRole(Owner) => new AgoraPermissions(Read, Write, Create, Redact, Manage)
+      case GcsBucketRole(Writer) => new AgoraPermissions(Read, Write, Create, Redact)
+      case GcsBucketRole(Reader) => AgoraPermissions(Read)
+      case _ => AgoraPermissions(AgoraPermissions.Nothing) 
+    }
+  }
+
+  def gcsObjectToMethodPermissions(objectRole: GcsObjectRole): AgoraPermissions = {
+    objectRole match {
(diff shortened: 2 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.broadinstitute.dsde.agora.server.business.AgoraPermissions._
+import org.broadinstitute.dsde.agora.server.acls.GcsRole._
+
+/**
+ * Created by dshiga on 6/16/15.
+ */
+object RoleTranslator {
+
+  def gcsBucketToNamespacePermissions(bucketRole: GcsBucketRole): AgoraPermissions = {
+    var namespacePermissions = new AgoraPermissions(AgoraPermissions.Nothing)
```
bradtaylor: Can this be done without a Var?  Perhaps doing a bucketRole.Match statement?

src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
```diff
@@ -0,0 +1,61 @@
...
+    if (bucketRole.isReader) {
+      namespacePermissions = namespacePermissions.addPermissions(Read)
+    }
+    if (bucketRole.isWriter) {
+      namespacePermissions = namespacePermissions.addPermissions(Write)
```
bradtaylor: can't we say namespacePermissions.addPermission(Write, Create, Redact)?  I thought Jay implemented vararg constructor for AgoraPermissions

src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
```diff
@@ -0,0 +1,61 @@
...
+      namespacePermissions = namespacePermissions.addPermissions(Redact)
+    }
+    if (bucketRole.isOwner) {
+      namespacePermissions = namespacePermissions.addPermissions(Manage)
+    }
+    namespacePermissions
+  }
+
+  def gcsObjectToMethodPermissions(objectRole: GcsObjectRole): AgoraPermissions = {
+    var methodPermissions = new AgoraPermissions(AgoraPermissions.Nothing)
```
bradtaylor: Similar question of whether var can be avoided

src/test/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslatorTest.scala
```diff
@@ -0,0 +1,102 @@
...
+    val objectRole = GcsObjectRole("READER")
+    val methodPermissions = RoleTranslator.gcsObjectToMethodPermissions(objectRole)
+    assert(methodPermissions.canRead === true)
+    assert(methodPermissions.canWrite === false)
+    assert(methodPermissions.canRedact === false)
+    assert(methodPermissions.canManage === false)
+  }
+
+  "RoleTranslator" should "translate object OWNER -> method read, manage" in {
+    val bucketRole = GcsBucketRole("NOTHING")
```
bradtaylor: is bucketRole being used for anything in this test?

src/main/scala/org/broadinstitute/dsde/agora/server/acls/GcsRole.scala
```diff
@@ -0,0 +1,30 @@
+package org.broadinstitute.dsde.agora.server.acls
+
+import org.broadinstitute.dsde.agora.server.acls.GcsRole._
+
+/**
+ * Created by dshiga on 6/16/15.
+ */
+
```
jacarey: We don't need author tags. All of the information we need about who had done what is in github and at a much more granular level. This is probably code template in your Intellij for new classes that you can just set to blank.

jacarey:
Looks good.

src/main/scala/org/broadinstitute/dsde/agora/server/acls/GcsRole.scala
geoffjentry: What about rewriting history via rebase, you insensitive clod! :)

src/main/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslator.scala
dshiga: Good point, changing.

src/test/scala/org/broadinstitute/dsde/agora/server/acls/RoleTranslatorTest.scala
dshiga: No, good catch.

PR #69, opened 2015-06-22
Restrict inserts to users with create permission

dshiga:
Also, move authorization checks up from the business layer to the add and query handlers. Benefits of this are:
-We don't have to pass an AuthorizationProvider all the way down to the business layer
-We don't have to throw an exception in the business layer when a user is not authorized to insert

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -5,7 +5,7 @@ import org.broadinstitute.dsde.agora.server.AgoraConfig
 import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
 import org.broadinstitute.dsde.agora.server.model.{AgoraEntity, AgoraEntityProjection, AgoraEntityType}
 
-class AgoraBusiness(authorizationProvider: AuthorizationProvider) {
+class AgoraBusiness {
   def agoraUrl(entity: AgoraEntity): String = {
     hasNamespaceNameId(entity) match {
       case true => AgoraConfig.urlFromType(entity.entityType) + entity.namespace.get + "/" + entity.name.get + "/" + entity.snapshotId.get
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -5,7 +5,7 @@ import org.broadinstitute.dsde.agora.server.AgoraConfig
 import org.broadinstitute.dsde.agora.server.dataaccess.AgoraDao
 import org.broadinstitute.dsde.agora.server.model.{AgoraEntity, AgoraEntityProjection, AgoraEntityType}
 
-class AgoraBusiness(authorizationProvider: AuthorizationProvider) {
+class AgoraBusiness {
```
bradtaylor: Is there still value in having this layer, if all it seems to add over the DAO is the URL?  More a question for George/Jay when they get back. Not a problem for this pull I think

src/main/scala/org/broadinstitute/dsde/agora/server/business/AuthorizedAgoraEntity.scala
```diff
@@ -3,4 +3,4 @@ package org.broadinstitute.dsde.agora.server.business
 
 import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 
-case class AuthorizedAgoraEntity(entity: AgoraEntity, authorization: AgoraPermissions)
+case class AuthorizedAgoraEntity(entity: Option[AgoraEntity], authorization: AgoraPermissions)
```
bradtaylor: Why is this becoming Option? I probably just don't see it

dshiga: Good question - I probably should have explained this. See QueryHandler line 50. When trying to get a single method by namespace/name/snapshotId, we get Some(Entity) or None in response. Then we have to check for permissions, which involves creating an AuthorizedAgoraEntity. I didn't want to have to figure out whether the response was Some or None before doing the auth check.

bradtaylor:
Thumbsup from me. We should have the actual add/insert call a function to create the google bucket/object. I think this belongs to the work Jay/George/I am doing on a separate branch

PR #71, opened 2015-06-26
Working version of method read authorization restriction

bradtaylor:
GcsAuthorizaitonProvider is making spray-client calls to GCS JSON API endpoints. 404 responses for ACL get are interpreted as Nothing permission.

For the method/config list endpoint, we collect the http response futures into a single future of a sequence, and await the result of that.
This allows futures to complete in parallel, ensuring we only block until the slowest http response arrives.

Bubbling GCS contact failures up to the query-handler as a ClientServiceFailure exception, which the queryHandler catches. Responds to user as 500 error.

Moved a ton of stuff into new packages

This version works on pre-created buckets.
Not sure this is the best way to handle the refresh token when authenticating GCS contact


src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/AuthorizedAgoraEntity.scala
```diff
@@ -1,6 +1,8 @@
 
-package org.broadinstitute.dsde.agora.server.business
+package org.broadinstitute.dsde.agora.server.dataaccess.acls
 
 import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 
 case class AuthorizedAgoraEntity(entity: Option[AgoraEntity], authorization: AgoraPermissions)
+
+case class AuthorizedNamepace(namepace: Option[String], authorization: AgoraPermissions)
```
jacarey: namepace -> namespace

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/gcs/GcsAuthorizationProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  def namespaceToBucketName(namespace: String): String = "agoraNamespace_" + namespace
+
+  def entityToObjectName(entity: AgoraEntity): String = "agoraEntity_" + entity.entityType.get + "_" + entity.name.get + "_" + entity.snapshotId.get
+
+  def getGcsObjectRoleRequestUrl(agoraEntity: AgoraEntity, userEmail: String): String = {
+    val agoraBusiness = new AgoraBusiness
+    agoraBusiness.hasNamespaceNameId(agoraEntity) match {
+      case true =>
+        "https://www.googleapis.com/storage/v1/b/" + namespaceToBucketName(agoraEntity.namespace.get) + "/o/" + entityToObjectName(agoraEntity) + "/acl/user-" + userEmail
```
jacarey: lets yank out the google api url and put it in a constant. When the rev the api it will make changing it easier.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/gcs/GcsAuthorizationProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+  override def authorizationsForEntities(agoraEntities: Seq[AgoraEntity], username: String): Seq[AuthorizedAgoraEntity] = {
+    // Get a sequence of response Futures, one for each Some(entity). Package the http response alongside the entity.
+    // Transform the sequence of Futures into a Future of a sequence
+    val responseFutures: Future[Seq[GoogleResponse]] = Future.sequence(agoraEntities.map(entity => {
+      val httpResponseFuture = pipeline(Get(getGcsObjectRoleRequestUrl(entity, username)))
+      httpResponseFuture.map(httpResponse => new GoogleResponse(entity, httpResponse))
+    }))
+
+    // Futures can complete in parallel, and we await the result on the slowest HttpResponse
+    val googleResponses: Seq[GoogleResponse] = Await.result(responseFutures, Duration.Inf)
```
jacarey: Do we really want to wait forever or should we add a reasonable timeout? What happens if we get a timeout?

src/main/scala/org/broadinstitute/dsde/agora/server/model/GoogleAccessControls.scala
```diff
@@ -0,0 +1,85 @@
...
+                          owner: Option[Owner] = None,
+                          location: Option[String] = None,
+                          website: Option[Website] = None,
+                          logging: Option[Logging] = None,
+                          versioning: Option[Versioning] = None,
+                          cors: List[Cors] = Nil,
+                          lifecycle: Option[Lifecycle] = None,
+                          storageClass: Option[String] = None,
+                          etag: Option[String] = None
+                           )
```
jacarey: mixed ( placing.  Do we like putting it on the line with the first param or a separate line? Personally I am starting to prefer the same line but we should decide and be consistent.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -45,10 +46,10 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
 
   private def add(requestContext: RequestContext, agoraEntity: AgoraEntity, username: String): Unit = {
-    val authorizedEntity = authorizationProvider.authorizationsForEntity(Some(agoraEntity), username)
-    if (!authorizedEntity.authorization.canCreate) {
-      context.parent ! RequestComplete(BadRequest, AgoraError("You don't have permission to create items in " + agoraEntity.namespace))
-    }
+//    val authorizedEntity = authorizationProvider.authorizationsForEntity(Some(agoraEntity), username)
+//    if (!authorizedEntity.authorization.canCreate) {
+//      context.parent ! RequestComplete(BadRequest, AgoraError("You don't have permission to create items in " + agoraEntity.namespace))
+//    }
```
jacarey: Why comment and not delete?

bradtaylor: This supports the write, and I'm going to roll right into working on that. I just want to leave this here for myself as a TODO hook to come back to

jacarey: Gotcha

src/main/resources/application.conf
```diff
@@ -0,0 +1,13 @@
...
+
+openam {
+  deploymentUri = "https://openam-ci.vault.broadinstitute.org/opensso"
+  realm = "/apollo"
+}
+
+// Agora needs an associated google-project under which to create GCS objects.
+// Create a service account with p12 key associated with this project and record it's info in application.conf on the deployment server.
+gcs.project.id = "broad-dsde-dev"
+gcs.service.account.email = "806222273987-320bupcp6q0ekab4vobolvpb1oec8c3s@developer.gserviceaccount.com"
```
jacarey: do we want this values in source control?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
```diff
@@ -59,8 +68,14 @@ class QueryHandler(authorizationProvider: AuthorizationProvider) extends Actor {
             entityTypes: Seq[AgoraEntityType.EntityType],
             username: String): Unit = {
     val entities = agoraBusiness.find(agoraSearch, agoraProjection, entityTypes, username)
-    authorizationProvider.filterByReadPermissions(authorizationProvider.authorizationsForEntities(entities, username))
-    context.parent ! RequestComplete(entities)
+    try {
+      val authorizedEntities = authorizationProvider.filterByReadPermissions(authorizationProvider.authorizationsForEntities(entities, username))
```
jacarey: making two subsequent calls to the auth provider seems a bit silly. Maybe we should create a few helper methods like
filterEntitiesByReadPermissions(agoraEntities: Seq[AgoraEntity], username: String)
then we can compose this line like
context.parent ! RequestComplete(filterEntitiesByReadPermissions(entities, username)

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
```diff
@@ -59,8 +68,14 @@ class QueryHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
+      context.parent ! RequestComplete(authorizedEntities)
+    }
+    catch {
+      case clientServiceFailure: ClientServiceFailure => context.parent ! RequestComplete(InternalServerError, "Error encountered in entity authorization")
+      case timeoutException: TimeoutException => context.parent! RequestComplete(InternalServerError, "Timeout exception encountered during entity authorization")
+    }
```
jacarey: Identical catch clauses makes me think we could refactor this a bit.

src/main/scala/org/broadinstitute/dsde/agora/server/model/GoogleAccessControls.scala
jacarey: I think as long as the parameters fit on a single 120 column line we don't need to separate them all.

Otherwise they should be broken out like

case class BucketResource(name: Option[String] = None,
                          timeCreated: Option[String] = None, 
                          metageneration: Option[Long] = None,
                          acl: List[String] = Nil,
                          defaultObjectAcl: List[String] = Nil,
                          owner: Option[Owner] = None,
                          location: Option[String] = None,
                          website: Option[Website] = None,
                          logging: Option[Logging] = None,
                          versioning: Option[Versioning] = None,
                          cors: List[Cors] = Nil,
                          lifecycle: Option[Lifecycle] = None,
                          storageClass: Option[String] = None,
                          etag: Option[String] = None)

Does that seem reasonable to everyone?

src/main/resources/application.conf
bradtaylor: UGHGHGHGHG No. This got in via a bad commit. I'll rebase it out.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
bradtaylor: Yes that bothered me as well.

src/main/scala/org/broadinstitute/dsde/agora/server/model/GoogleAccessControls.scala
bradtaylor: Agreed. Will fix everywhere in the codebase

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
bradtaylor: Jay and I talked in person about various ways to refactor this. None were attractive. Leaving as-is. We can circle back and clean it up when implementing write or edit if people have any ideas

PR #75, opened 2015-07-01
added an only payload endpoint for execution service

vivster7:
Changed queryByNamespaceNameSnapshotId to querySingle. 

I tried to minimize the amount of logic in the route itself. Personally, I think this is more readable, and we should refactor the other routes. Let me know what you think.

Also added some file level comments -- stuff I would have found helpful when I first dove into the codebase.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
@@ -1,14 +1,25 @@
 
 package org.broadinstitute.dsde.agora.server.webservice
 
+
+
+import scala.util.Try
 import akka.actor.Props
 import org.broadinstitute.dsde.agora.server.dataaccess.acls.AuthorizationProvider
 import org.broadinstitute.dsde.agora.server.model.{AgoraEntity, AgoraEntityProjection, AgoraEntityType}
 import org.broadinstitute.dsde.agora.server.webservice.util.ServiceMessages
 import org.broadinstitute.dsde.agora.server.webservice.validation.{AgoraValidation, AgoraValidationRejection}
 import org.joda.time.DateTime
-import spray.routing.HttpService
+import spray.routing.{RequestContext, HttpService}
 
+
+/**
+ * AgoraService defines routes for ApiServiceActor.
+ *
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -17,32 +28,51 @@ abstract class AgoraService(authorizationProvider: AuthorizationProvider) extend
...
       }
     }
 
+  // Route helpers
+    val matchPath = get & path(path / Segment / Segment / IntNumber)
+    val extractOnlyParameter = extract(_.request.uri.query.get("onlyPayload"))
+
+    def extractBool(x: Option[String]): Boolean = {
+      x match {
+        case Some(x) => Try(x.toBoolean).getOrElse(false)
```
bradtaylor: Now that I get a look at this, very good use of Try. Love it.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -17,32 +28,51 @@ abstract class AgoraService(authorizationProvider: AuthorizationProvider) extend
...
+        case None => false
+      }
+    }
```
jacarey: extract to util

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -17,32 +28,51 @@ abstract class AgoraService(authorizationProvider: AuthorizationProvider) extend
...
-        usernameFromCookie() { userName =>
+  // Route: GET http://root.com/<namespace>/<name>/<snapshotId>
+  // Route: GET http://root.com/<namespace>/<name>/<snapshotId>?onlyPayload=True
+  def querySingleRoute =
+
+  // Route logic
+    matchPath { (namespace, name, snapshotId) =>
+      usernameFromCookie() { (username) =>
+        extractOnlyParameter { (only) =>
+          val onlyBool = extractBool(only)
```
jacarey: don't need val just call extract below

jacarey:
Looks good.

bradtaylor:
Thumbs up!

PR #76, opened 2015-07-01
Ignore cromwell's logback.xml to avoid sbt assembly error

dshiga:
We were getting errors when doing sbt assembly because the latest cromwell snapshot jar released 6/30 includes a logback.xml file, which causes problems because agora has its own logback.xml too.


vivster7:
looks fine for merge

jacarey:
:+1:  It would be nice if cromwell had a library jar that wasn't their full assembled jar.

nh13:
@jacarey add a request into the red team's backlog

PR #77, opened 2015-07-07
Fix embedded urls

dshiga:
Our embedded urls had port 8000 in them, which is not actually what we want when agora runs behind a proxy. When we run behind a proxy we usually want to leave the port blank in the embedded url.

I created a new config parameter called embeddedUrl.port so that this can be set independently of whatever port we tell agora to bind to. So, for example we can have:
webservice.port=8000
embeddedUrl.port=443

Or we can just not specify embeddedUrl.port, in which case the port will be blank in the embedded url.

We still need the option of specifying a port in embedded urls, because people may sometimes run locally without a proxy and still want to have valid embedded urls.


bradtaylor:
Looks good

PR #79, opened 2015-07-10
Supply an import resolver when validating wdl

dshiga:
The cromwell jar requires us to supply a wdl resolver function when we ask it to validate wdl. We were getting away with not supplying one before because the default resolver was a no-op, but that has changed.

This pull request creates a wdl resolver function. Given a uri of the form "methods://namespace.name.snapshotId", it finds the method and returns its payload. If it can't find the method for any reason it throws an exception.


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -53,4 +53,69 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
   }
+
+}
+
+case class MethodImportResolver(username: String, business: AgoraBusiness, authorizationProvider: AuthorizationProvider) {
+
+  def importResolver(importUri: String): WdlSource = {
+    ImportResolverHelper.validateUri(importUri)
+    val method = ImportResolverHelper.resolve(importUri, business, username)
+    val authorizedMethod = authorizationProvider.filterByReadPermissions(method, username)
```
jacarey: Is there not a "has read permission" method for a single entity? Seems weird to be filtering on a single entity instead of doing a boolean check.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -53,4 +53,69 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
+    authorizedMethod match {
+      case x: Some[AgoraEntity] => x.get.payload.get
```
jacarey: why not 
case Some(entity) => entity.payload.get
?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -53,4 +53,69 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
+    }
+  }
+}
+
+object ImportResolverHelper {
+
+  private val methodScheme = "methods://"
+
+  def resolve(uri: String, business: AgoraBusiness, username: String): Option[AgoraEntity] = {
+    val withoutScheme = removeScheme(uri)
```
jacarey: Would it make sense to create a case class that can take a uri and do all of the parsing internally so that we can use it elsewhere if we need to?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -53,4 +53,69 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
+
+  def validateUri(importUri: String) = {
+    if (!importUri.startsWith(methodScheme)) {
+      throw new SyntaxError(s"Unsupported import uri: $importUri. Must start with $methodScheme.")
+    }
+    val withoutScheme = removeScheme(importUri)
+    if (!hasThreeParts(withoutScheme)) {
+      throw new SyntaxError(s"Malformed methods uri. Must contain three parts, separated by '.': $importUri")
+    }
+    val parts = importUri.split("\\.")
```
jacarey: this validation method splits that string twice

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -53,4 +53,69 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
+    parts.size == 3
+  }
+
+  private def isIdInt(idStr: String): Boolean = {
+    try {
+      idStr.toInt
+      true
+    } catch {
+      case e: NumberFormatException => false
+    }
```
jacarey: why not idStr.forall(_.isDigit)? Might be better than parsing the string to an int that we are just throwing away

jacarey:
Looks good

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
vivster7: There is in the new GcsAuthProvider. It's called isAuthorizedForRead. It's currently in review though.

dshiga: Maybe we should hold off merging this until after the new GcsAuthProvider gets merged. We're going to get conflicts either way, but probably makes more sense for me to adapt to the new auth provider in my branch.

vivster7: Thanks for waiting dave. With the new GcsAuthorizationProvider, I think this code should look like

```scala
import org.broadinstitute.dsde.agora.server.dataaccess.acls.gcs.GcsAuthorizationProvider._
if (isAuthorizedForRead(method, username)) method.payload.get
else throw new SyntaxError(s"Can't resolve import: $importUri")
```

Now that I write this, I think the variable name "entity" might be better than "method"

dshiga: Hmm. Imports only apply to methods though - we won't ever have imports to resolve for configurations as far as I know. So this val will always be a method.

vivster7: ahh, okay, then method is a better name.

PR #80, opened 2015-07-10
google credential token now refreshes when it expires (every hour)

vivster7:
Quick fix to make the GoogleCredential token refresh when it expires. Pipeline has been changed to a function so the token will be checked before every call to Google.


src/main/scala/org/broadinstitute/dsde/agora/server/GoogleCredentialHandler.scala
```diff
@@ -9,18 +9,29 @@ import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.storage.StorageScopes.DEVSTORAGE_FULL_CONTROL
 
 object GoogleCredentialHandler {
-  val accessToken: String = {
+
+  val one_minute = 60
```
jacarey: variable name should be oneMinute alternately use scala.concurrent.duration 1.minute instead of the val

jacarey:
:+1:

PR #81, opened 2015-07-13
Fix unmarshalling when OpenAm user's email is empty

dshiga:
Agora was blowing up when requests were coming from a OpenAM test user with no email set. We may want to block such users, but we shouldn't do it by blowing up. This pull request adds a custom json format that can deal with the empty email field.


jacarey:
Looks good. I'm pretty sure the build failure is just that random timeout we get in travis these days when we bind port at startup.

PR #84, opened 2015-07-15
Moved all business logic in to AgoraBusiness (including authorization)

jacarey:
Added stub suite for integration testing.
Added authorization exception that is handled by the supervisor of PerRequest.


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraAuthorizationException.scala
```diff
@@ -0,0 +1,16 @@
+
+package org.broadinstitute.dsde.agora.server.business
+
+import org.broadinstitute.dsde.agora.server.dataaccess.acls.AgoraPermissions
+import org.broadinstitute.dsde.agora.server.model.AgoraEntity
+
+case class AgoraAuthorizationException(agoraPermissions: AgoraPermissions, agoraEntity: AgoraEntity) extends Exception {
+  override def getMessage: String = {
+    "Authorization exception attempting to exercise permission %d on entity %s.%s.%d. %s".format(
```
bradtaylor: Should we prefer string interpolation to format?
s"Authorization exception attempting to exercise permission $agoraPermissions.permissions

It would make this string longer, but you could break it up over multiple lines with +

Up to you which is more readable

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -33,85 +34,7 @@ class AddHandler(authorizationProvider: AuthorizationProvider) extends Actor {
...
-    uri.substring(methodScheme.length)
-  }
-
-  private def isIdInt(idStr: String): Boolean = {
-    idStr.forall(_.isDigit)
-  }
-
-  private def uriParts(uriWithoutScheme: String): Array[String] = {
-    uriWithoutScheme.split("\\.")
+    context.parent ! RequestComplete(Created, agoraBusiness.insert(agoraEntity.copy(createDate = Option(new DateTime())), username))
```
bradtaylor: AgoraBusiness.insert performs the entity copy to add createDate. See
https://github.com/broadinstitute/agora/pull/84/files#diff-1768f9d51119592601a48548da8f9904R16

Remove the copy here

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/PerRequest.scala
```diff
@@ -53,6 +54,10 @@ trait PerRequest extends Actor {
 
   override val supervisorStrategy =
     OneForOneStrategy() {
+      case authException: AgoraAuthorizationException =>
+        system.log.error(authException, "Authorization exception processing request:" + r.request.uri)
```
bradtaylor: error?  Or info?  It's not really a system error that a user tried to foo a bar without access. It's just dumbfoolery on their part

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
```diff
@@ -14,29 +12,31 @@ import spray.httpx.SprayJsonSupport._
...
 
   def receive = {
     case QuerySingle(requestContext: RequestContext,
-                     namespace: String,
-                     name: String,
-                     snapshotId: Int,
-                     entityType: Seq[AgoraEntityType.EntityType],
-                     username: String,
-                     onlyPayload: Boolean) =>
+    namespace: String,
```
bradtaylor: Why was the indenting removed here. I thought we decided we liked that? See every case class under Model. But this is a case match of course. Does intelliJ complain or something?

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/authorization/TestAuthorizationProvider.scala
```diff
@@ -12,8 +13,8 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 object TestAuthorizationProvider extends AuthorizationProvider with AgoraTestData{
 
   // local permission storage.
-  val entityPermissions = new HashMap[String, AgoraPermissions]()
-  val namespacePermissions = new HashMap[String, AgoraPermissions]()
+  val entityPermissions = new mutable.HashMap[String, AgoraPermissions]()
```
bradtaylor: Why did this need to become a mutable map?

bradtaylor:
Several comments. Otherwise thumbsup

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/authorization/TestAuthorizationProvider.scala
jacarey: It always was. This is just a scala convention for calling out mutable collections in the instantiation. (A mild form of shaming, but mostly just to make it obvious you are dealing with mutable collection without looking at the imports)

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/PerRequest.scala
jacarey: info sounds good

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
jacarey: will do

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraAuthorizationException.scala
jacarey: I think string interpolation makes it easier to read what the string is meant to represent. I'll change it.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
jacarey: Grrr stupid auto formatter does this for case classes. I'll fix it

jacarey:
I think it is late enough in this sprint to delay pushing this to master to avoid any pre-demo complications.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
jacarey: FYI - Checking "Align when multiline" for Method call arguments in File -> Settings... -> Code Style -> Scala on the Wrapping and Braces tab fixes this.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -42,40 +38,40 @@ class AgoraBusiness {
                  username: String): Option[AgoraEntity] = {
     val optEntity = AgoraDao.createAgoraDao(entityTypes).findSingle(namespace, name, snapshotId)
     optEntity match {
-      case Some(foundEntity) => Some(foundEntity.copy(url = Option(agoraUrl(foundEntity))))
+      case Some(foundEntity) =>
```
vivster7: Instead of returning an Option, we should just throw the error to be caught by Supervisor. Will also save some lines of code in QueryHandler

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AddHandler.scala
```diff
@@ -15,17 +15,18 @@ import spray.httpx.SprayJsonSupport._
 import spray.routing.RequestContext
 
 /**
- * Handles adding a method to the methods repository, including validation.
+ * AddHandler is an actor that receives web service requests and calls AgoraBusiness logic.
+ * It then handles the returns from the business layer and completes the request. It is responsible for adding a method
+ * or method configuration to the methods repository, including validation.
```
vivster7: addHandler no longer handles validation. I believe it moved to AgoraBusiness

vivster7:
:+1:

PR #85, opened 2015-07-20
Route + Class validation ( + route refactor)

vivster7:


bradtaylor:
Could the 2 commits be squashed?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -20,35 +21,95 @@ object AgoraEntityType extends Enumeration {
...
+    def validateSnapshotId(_id: Int): ValidationNel[String, Int] = {
+      if (_id > 0) _id.successNel[String]
+      else "SnapshotId must be greater than 0".failureNel[Int]
+    }
+
+    def validateSynopsis(synopsis: String): ValidationNel[String, String] = {
+      if (synopsis.length <= 80) synopsis.successNel[String]
+      else "Synopsis must be less than 80 chars".failureNel[String]
+    }
+
```
jacarey: I believe there is a 10k documentation limit as well that we should validate.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/QueryHandler.scala
```diff
@@ -23,13 +23,11 @@ class QueryHandler(authorizationProvider: AuthorizationProvider) extends Actor {
 
   def receive = {
     case QuerySingle(requestContext: RequestContext,
-                     namespace: String,
-                     name: String,
-                     snapshotId: Int,
+                     entity: AgoraEntity,
                      entityType: Seq[AgoraEntityType.EntityType],
```
jacarey: Seq should be plural entityTypes

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
```diff
@@ -0,0 +1,135 @@
...
+               s"You can't perform operation for entity type $entityType.get at path /$path.")
+    else
+      pass
+  }
+
+  def toBool(x: Option[String]): Boolean = {
+    x match {
+      case Some(x) => Try(x.toBoolean).getOrElse(false)
+      case None => false
+    }
```
jacarey: i think this can be simplified to 

Try(x.get.toBoolean).getOrElse(false)

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessTest.scala
```diff
@@ -22,11 +22,11 @@ class AgoraBusinessTest extends FlatSpec with Matchers with AgoraTestData {
...
-
-  "Agora" should "not find a method payload when resolving a WDL import statement if the method has not been added" in {
-    val importString = "methods://broad.nonexistent.5400"
-    assert(agoraBusiness.importResolver(importString, agoraCIOwner.get) === "")
-  }
+//
+//  "Agora" should "not find a method payload when resolving a WDL import statement if the method has not been added" in {
+//    val importString = "methods://broad.nonexistent.5400"
+//    assert(agoraBusiness.importResolver(importString, agoraCIOwner.get) === "")
+//  }
```
jacarey: if this is a useful test that is currently not working we should set it to ignore instead of commenting it out so we don't lose track of it.

jacarey:
:+1:

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
```diff
@@ -0,0 +1,135 @@
...
+
+
+trait RouteHelpers extends AgoraDirectives
+  with QuerySingleRoute
+  with QueryRoute
+  with AddRoute
+
+trait BaseRoute extends PerRequestCreator with RouteUtil
+
+trait QuerySingleRoute extends BaseRoute {
```
bradtaylor: I'd still prefer this be called something like QuerySingleHelpers, since the actual route is a def in AgoraService. Just my preference.

vivster7: Hmm, personally, I think there is enough context: the file and parent traits are both called RouteHelpers. And the trait keyword suggests that you can find the concrete implementation elsewhere. What do other people think? @dshiga @gbggrant @jacarey

dshiga: I think QuerySingleHelper is clearer. To me the routes are just the url patterns that we match in AgoraService.

vivster7: sounds good, I'll shall change the name.

PR #89, opened 2015-07-23
Make base url configurable for CLI

dshiga:
The agora URL was hard coded to agora-ci.broadinstitute.org. I've defaulted it to agora-prod.broadinstitute.org, since that is what the orchestration service is using. I've also made it overridable, so you can pass in whatever you want, with -u.


jacarey:
+1

jacarey:
:+1:

PR #90, opened 2015-07-23
Added config file env variable.

jacarey:
If TEST_CONFIG is set in your environment it will be used when sbt runs its tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -81,18 +81,22 @@ val customMergeStrategy: String => MergeStrategy = {
   case _ => MergeStrategy.deduplicate
 }
 
-mergeStrategy in assembly := customMergeStrategy
-
-test in assembly := {}
-
 val integrationTest = inputKey[Unit]("Runs the Agora integration tests.")
+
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -81,18 +81,22 @@ val customMergeStrategy: String => MergeStrategy = {
...
   (testOnly in Test).toTask(" org.broadinstitute.dsde.agora.server.AgoraUnitTestSuite").value
-}
\ No newline at end of file
+}
+
+mergeStrategy in assembly := customMergeStrategy
+
+fork in Test:= true
+
+javaOptions in Test ++= Seq("-Dconfig.file=" + Option(System.getenv("TEST_CONFIG")).getOrElse("src/test/resources/reference.conf"))
```
dshiga: Should we just error out if we can't find the proper test config file?

jacarey: this is mainly just to allow us to inject another conf file... otherwise it just falls back on the previous behavior

dshiga: Ok, :+1:

PR #93, opened 2015-07-24
Let users specify synopsis as parameter to CLI

dshiga:
Previously, trying to push a method with the CLI would open an editor, requiring the user to type in a synopsis for every push. Now you can optionally specify it as a parameter, which makes for less typing.


vivster7:
:+1:

PR #94, opened 2015-07-27
Improvements to agora testing

gbggrant:
Load and validate environment (local, dev, ci, prod) from typesafe config

conditionally use embeddedMongo based on environment (only if local)
Refactored to move MockAuthorizationProvider (nee TestAuthorizationProvider) from test to src.
Select gcsAuthorizationProvider or mockAuthorizationProvider based on environment (i.e. local uses mock).
Passed environment to Agora object so that unitTest and integrationTest could use different environments for their configuration.
Pushed env-specific initialization down to ServerInitializer and ApiServiceActor.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -7,6 +7,30 @@ instance.name = "reference"
 mongodb.host = "localhost"
 mongodb.port = 27017
 
+environment = "local"
+prod {
+  useOpenAMAuthentication = true
+  useStandaloneMongo = true
+  useGcsAuthorizationProvider = true
+}
+ci {
+  useOpenAMAuthentication = true
+  useStandaloneMongo = true
+  useGcsAuthorizationProvider = true
+}
+dev {
+  useOpenAMAuthentication = true
+  useStandaloneMongo = true
+  useGcsAuthorizationProvider = true
+}
+local {
+  useOpenAMAuthentication = false
+  useStandaloneMongo = false
+  useGcsAuthorizationProvider = false
+  // If not using openAMAuthentication, you need to supply a mock authenticated email.
+  mockAuthenticatedUserEmail = "broadprometheustest@gmail.com"
+}
+
 swagger {
   apiDocs = "api-docs"
   apiVersion = "0.1"
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
@@ -8,6 +8,23 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntityType.EntityType
 
 object AgoraConfig {
   private val config: Config = ConfigFactory.load()
+  
+  val LocalEnvironment = "local"
+  val DevEnvironment = "dev"
+  val CiEnvironment = "ci"
+  val ProdEnvironment = "prod"
+
+  val Environments = List[String](LocalEnvironment, DevEnvironment, CiEnvironment, ProdEnvironment)
+
+  lazy val environment = config.as[Option[String]]("environment").getOrElse(LocalEnvironment)
+  if (!Environments.contains(AgoraConfig.environment))
+    throw new IllegalArgumentException("Illegal environment '" + AgoraConfig.environment + "' specified.")
+
+  val envConfig = config.getConfig(environment)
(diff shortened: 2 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/MockAgoraDirectives.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.broadinstitute.dsde.vault.common.directives.OpenAMDirectives
+import org.broadinstitute.dsde.vault.common.util.ImplicitMagnet
+import spray.routing.Directives._
+import spray.routing._
+
+import scala.concurrent.ExecutionContext
+
+trait MockAgoraDirectives extends AgoraDirectives {
+  def commonNameFromCookie(magnet: ImplicitMagnet[ExecutionContext]): Directive1[String] = {
+    OpenAMDirectives.commonNameFromCookie(magnet)
```
dshiga: I'm confused about why we are still getting the common name from OpenAM here - should we be grabbing a name from AgoraConfig instead?

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -8,6 +8,24 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntityType.EntityType
...
   private val config: Config = ConfigFactory.load()
+  
+  val LocalEnvironment = "local"
+  val DevEnvironment = "dev"
+  val CiEnvironment = "ci"
+  val ProdEnvironment = "prod"
+
+  val Environments = List[String](LocalEnvironment, DevEnvironment, CiEnvironment, ProdEnvironment)
+
+  // TODO - what is the default??
```
dshiga: Can we remove this comment now? I think we decided that local would be the default, which is what you're doing.

src/main/resources/reference.conf
```diff
@@ -7,6 +7,30 @@ instance.name = "reference"
...
+  useOpenAMAuthentication = true
+  useStandaloneMongo = true
+  useGcsAuthorizationProvider = true
+}
+local {
+  useOpenAMAuthentication = false
+  useStandaloneMongo = false
+  useGcsAuthorizationProvider = false
+  // If not using openAMAuthentication, you need to supply a mock authenticated email.
+  mockAuthenticatedUserEmail = " broadprometheustest@gmail.com"
```
dshiga: extra space in email above

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -8,6 +8,24 @@ import org.broadinstitute.dsde.agora.server.model.AgoraEntityType.EntityType
...
+  lazy val environment = config.as[Option[String]]("environment").getOrElse(LocalEnvironment)
+  if (!Environments.contains(AgoraConfig.environment))
+    throw new IllegalArgumentException("Illegal environment '" + AgoraConfig.environment + "' specified.")
+
+  val envConfig = config.getConfig(environment)
+  val mockAuthenticatedUserEmail = envConfig.as[Option[String]]("mockAuthenticatedUserEmail").getOrElse("noone@broadinstitute.org")
+  def getUseOpenAMAuthenticationForEnvironment(environment: String) : Boolean = {
```
dshiga: Perhaps we could shorten / change to something like isUseOpenAmAuth? The "for environment" part is implied because environment is a parameter. Similarly, getUseGcsAuthorizationProviderForEnvironment could perhaps be isUseGcsAuth or something like that.

dshiga:
Made a few comments. Otherwise, :+1:

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/MockAgoraDirectives.scala
gbggrant: Yes, I think you're right.  commonNameFromCookie doesn't appear to be being used.

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
gbggrant: yes - thanks.

jacarey:
:+1:

PR #95, opened 2015-07-29
Added negative test case to GCS auth tests when a bucket exists but u…

jacarey:
…ser doesn't have write permissions.

Fixed potential None.get exception on AgoraAuthorizationException.

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/GcsAuthorizationSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/GcsAuthorizationSpec.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.agora.server.dataaccess.acls
 
 import org.broadinstitute.dsde.agora.server.AgoraConfig
 import org.broadinstitute.dsde.agora.server.AgoraIntegrationTestData._
-import org.broadinstitute.dsde.agora.server.business.AgoraBusiness
+import org.broadinstitute.dsde.agora.server.business.{AgoraAuthorizationException, AgoraBusiness}
 import org.broadinstitute.dsde.agora.server.dataaccess.acls.gcs.GcsAuthorizationProvider
 import org.scalatest.{DoNotDiscover, FlatSpec}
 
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/GcsAuthorizationSpec.scala
```diff
@@ -16,4 +16,13 @@ class GcsAuthorizationSpec extends FlatSpec {
     val entity = agoraBusiness.insert(testIntegrationEntity, "jcarey@broadinstitute.org")
     assert(agoraBusiness.findSingle(entity, Seq(entity.entityType.get), username) === entity)
   }
+
+  "Agora" should "check bucket write ACL when creating a new method in an existing workspace and fail if the user " +
+    "doesn't have access" in {
+    val thrown = intercept[AgoraAuthorizationException] {
+      val entity = agoraBusiness.insert(testIntegrationEntity, "ggrant@broadinstitute.org")
+      assert(agoraBusiness.findSingle(entity, Seq(entity.entityType.get), username) === entity)
```
dshiga: Do we need this line? The code should never reach here because an exception is thrown, right?

jacarey: good point... was from a copy paste.

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/GcsAuthorizationSpec.scala
```diff
@@ -16,4 +16,13 @@ class GcsAuthorizationSpec extends FlatSpec {
...
+    }
+    assert(thrown != null)
+  }
 }
```
gbggrant: This new test fails because of the order of execution - right?  Are these tests always guaranteed to executed in this order?

jacarey: Within a FlatSpec yes. they are sequential. Across suites.... potentially not.

PR #96, opened 2015-07-30
Rewrote entity insert acl logic; view and edit acls

vivster7:


src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
```diff
@@ -0,0 +1,122 @@
...
+      val bucketAcls = listBucketAcl(bucket).setFields("items(entity,role)").execute()
+
+      bucketAcls.toPrettyString
+
+    } else {
+      throw new NamespaceAuthorizationException(AgoraPermissions(Manage), entity)
+    }
+  }
+
+  def insertNamespaceAcl(_context: RequestContext, entity: AgoraEntity, user: String, acl: BucketAccessControl) = {
```
gbggrant: Why the '_' prefix on variable names in some cases and not others?  If not needed I wouldn't use them.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
```diff
@@ -0,0 +1,122 @@
...
+      val updatedAcls = filterBucketAclsByEntity(bucketAcls, acl.getEntity)
+
+      patchBucketAcls(bucket, updatedAcls).setFields("acl(entity,role)").execute().toPrettyString
+
+    } else {
+      throw new NamespaceAuthorizationException(AgoraPermissions(Manage), entity)
+    }
+  }
+
+  def filterBucketAclsByEntity(acls: java.util.List[BucketAccessControl], entityToFilter: String): java.util.ArrayList[BucketAccessControl] = {
```
gbggrant: I think this can be replaced with a scala filter expression.
Something like updatedAcls = acls.filter(_.getEntity != entityToFilter)

src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
```diff
@@ -0,0 +1,122 @@
...
+      val updatedAcls = filterObjectAclsByEntity(objectAcls, acl.getEntity)
+
+      patchObjectAcls(_object, updatedAcls).setFields("acl(entity,role)").execute().toPrettyString
+
+    } else {
+      throw new EntityAuthorizationException(AgoraPermissions(Manage), entity)
+    }
+  }
+
+  def filterObjectAclsByEntity(acls: java.util.List[ObjectAccessControl], entityToFilter: String): java.util.ArrayList[ObjectAccessControl] = {
```
gbggrant: Similar comment as above about replacing this with a filter expression.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/acls/gcs/GcsClient.scala
```diff
@@ -0,0 +1,151 @@
...
+  def createBucketAcl(bucket: Bucket, content: BucketAccessControl) =
+    bucketAcls.insert(bucket.getName, content)
+
+  def deleteBucketAcl(bucket: Bucket, user: String) =
+    bucketAcls.delete(bucket.getName, user.toGoogleUserEntity)
+
+  def listBucketAcl(bucket: Bucket) =
+    bucketAcls.list(bucket.getName)
+
+  def patchBuckettAcl(bucket: Bucket, user: String, content: BucketAccessControl) =
```
gbggrant: 'Bucket' (not 'Buckett')

src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
```diff
@@ -0,0 +1,122 @@
...
+    val numberOfAcls = acls.size
+    val updatedAcls = new java.util.ArrayList[ObjectAccessControl]()
+    for (i <- 0 until numberOfAcls) {
+      if (acls.get(i).getEntity != entityToFilter) {
+        updatedAcls.add(acls.get(i))
+      }
+    }
+    return updatedAcls
```
gbggrant: Unnecessary return statement.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AclHandler.scala
```diff
@@ -0,0 +1,48 @@
+package org.broadinstitute.dsde.agora.server.webservice
+
+import akka.actor.Actor
+import com.google.api.services.storage.model.{ObjectAccessControl, BucketAccessControl}
+import org.broadinstitute.dsde.agora.server.busines.AclBusiness
+import org.broadinstitute.dsde.agora.server.dataaccess.acls.AuthorizationProvider
+import org.broadinstitute.dsde.agora.server.model.{AgoraEntity}
```
gbggrant: Don't need braces '{' around single entry

gbggrant:
Looks good.  :+1:

src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
vivster7: typically, they are used to avoid naming conflicts (i.e, if there was another context variable in scope). In this case, there is not, so the _ can be removed.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AclBusiness.scala
vivster7: for this and the comment below, the google client api returns java ArrayList objects, which do not have filter functionality. The java arrays can be converted to scala arrays, so we can use filter, but they need to be converted back to Java in order to be returned to Google's API. I thought it was cleaner to just write the code as I would in Java instead of converting back and forth. I will add a comment to the code to make it clearer though.

PR #98, opened 2015-08-12
Do not do docker ref validation if docker image is on gcr.io

gbggrant:


bradtaylor:
Looks good, though at some future point we will want to validate gcr.io docker images, presumably.

How did this decrease our test coverage 17%?
Going to look at coveralls...

gbggrant:
Yes.  This was just added to bypass checking for the gcr.io docker images.  Just a stop gap to unblock Alex.

I don't understand the test coverage drop either.  I'll look a little deeper

PR #100, opened 2015-08-14
Add reference from Configuration to Method

dshiga:
When Agora gets a query for configurations, we need to return not only the configuration but also info about the associated method. This pull request implements that functionality.

Changes:
-When inserting a configuration, find the method reference in the payload and use it to populate the methodId metadata field in the configuration
-When returning configurations, use the methodId field to find the associated method and embed its details in the configuration json


src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/AgoraMongoDao.scala
```diff
@@ -55,7 +53,51 @@ class AgoraMongoDao(collections: Seq[MongoCollection]) extends AgoraDao {
     //insert the entity
     val dbEntityToInsert = EntityToMongoDbObject(entityWithId)
     collection.insert(dbEntityToInsert)
-    findSingle(entityWithId)
+    val insertedEntity = findSingle(MongoDbObjectToEntity(dbEntityToInsert))
+    insertedEntity
+  }
+
+  override def findSingle(namespace: String, name: String, snapshotId: Int): AgoraEntity = {
```
dshiga: I reordered some methods, so you'll see additions and deletions due purely to those moves. (I moved all the interface methods to the top).

PR #102, opened 2015-08-20
Refactored exceptions. Return 404 when user cannot read entity

vivster7:


src/main/scala/org/broadinstitute/dsde/agora/server/exceptions/AgoraException.scala
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsde.agora.server.exceptions
+
+import com.typesafe.scalalogging.slf4j.LazyLogging
+import spray.http.{StatusCode, StatusCodes}
+
+case class AgoraException(message: String = null,
+                          cause: Throwable = null,
+                          statusCode: StatusCode = StatusCodes.InternalServerError)
+  extends Exception(message, cause) with LazyLogging {
+  logger.debug("-"*20)
```
jacarey: this should be .error

jacarey:
:+1:

PR #108, opened 2015-08-28
Validate so user cannot post entity with snapshotId

vivster7:


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
```diff
@@ -180,7 +180,8 @@ trait AddRouteHelper extends BaseRoute {
 
   def validatePostRoute(entity: AgoraEntity, path: String): Directive0 = {
     validateEntityType(entity.entityType, path) &
-    validate(entity.payload.get.nonEmpty, "You must supply a payload.")
+    validate(entity.payload.get.nonEmpty, "You must supply a payload.") &
+    validate(entity.snapshotId.isEmpty, "You cannot specify a snapshotId. It will be assigned by the database.")
```
bradtaylor: I would say system rather than database, but it's nbd

bradtaylor:
thumbs up

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
vivster7: agreed

PR #111, opened 2015-09-01
seeking comments on changes fo google-based proxy instead of OpenAM (openidc)

kcibul:
Don't merge -- this needs to be done in a coordinated manner, but since Agora was the simplest service that uses OpenAM directly it seemed a good place to start.

added new OpenIdConnect directive to support google-based proxy, generified OpenAM names.


dshiga:
Looks reasonable to me.

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -5,7 +5,7 @@ import com.typesafe.config.{Config, ConfigFactory}
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.agora.server.model.AgoraEntityType
 import org.broadinstitute.dsde.agora.server.model.AgoraEntityType.EntityType
-import org.broadinstitute.dsde.agora.server.webservice.routes.{AgoraDirectives, AgoraOpenAMDirectives, MockAgoraDirectives}
+import org.broadinstitute.dsde.agora.server.webservice.routes.{AgoraDirectives, AgoraOpenAMDirectives, MockAgoraDirectives, OpenIdConnectDirectives}
 import scala.slick.jdbc.JdbcBackend.Database
```
jacarey: we should be able to get rid of the openam directives

jacarey:
:+1:

PR #112, opened 2015-09-01
refactor permissions client

vivster7:


vivster7:
tried to make this file easier to understand. woud like your opinion on how understandable you think it is. please point out any remaining areas of confusion

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -45,42 +45,55 @@ trait PermissionsClient {
...
-    Await.result(agoraPermissions, timeout)
   }
 
   def listPermissions(agoraEntity: AgoraEntity): Seq[AccessControl] = {
+    // Construct query
     val permissionsQuery = for {
       entity <- entities if entity.alias === alias(agoraEntity)
       _permissions <- permissions if _permissions.entityID === entity.id
       user <- users if user.id === _permissions.userID
     } yield (user.email, _permissions.roles)
```
gbggrant: Why not expand the query here like you have elsewhere (i.e. for permissionsQuery in getPermission)?

gbggrant:
I like it.  More readable than before.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
vivster7: honestly, its just a lack on Slick knowledge. This query is different because it returns a sequence of permissions, while the other queries return an individual permission. I would guess there is a way to expand / simplify the query, but I'm just not sure how to do it right now.

PR #114, opened 2015-09-03
added swagger endpoint for method redact

vivster7:


src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    "200": {
+                        "description": "An array of methods.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/MethodResponse"
+                            }
+                        }
+                    },
+                    "default": {
```
bradtaylor: Shouldn't this be "500"?  Currently its the only thing displayed under "Response Messages"

src/main/resources/swagger/agora.yaml
```diff
@@ -820,7 +885,7 @@ definitions:
       namespace:
```
bradtaylor: Should the name of this be MethodPost instead of method query?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    "201": {
+                        "description": "Method successfully added.",
+                        "schema": {
+                            "$ref": "#/definitions/MethodResponse"
+                        }
+                    },
+                    "400": {
+                        "description": "Malformed input.",
+                        "schema": {
+                            "$ref": "#/definitions/Error"
```
bradtaylor: The $ref here is not correctly associated with this response. Error is a 500 message, whereas this is 400. We need another definition for malformed response. Probably ones for 401 unauthorized and 404 not found as well

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    }
+                ],
+                "tags": [
+                    "Methods"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "An single method.",
+                        "schema": {
+                            "type": "array",
```
bradtaylor: Query single does not return an array

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                            "items": {
+                                "$ref": "#/definitions/MethodResponse"
+                            }
+                        }
+                    },
+                    "default": {
```
bradtaylor: 500 again?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                "tags": [
+                    "Methods"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "An array of methods.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/MethodResponse"
```
bradtaylor: We don't actually return some fields by default in the list (payload, documentation are excluded). Should we have a separate definition for MethodListResponse that makes this distinction clear?

bradtaylor: Is there any way to make it clear that payload and documentation fields are excluded by default?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    "Methods"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "An integer showing how many methods were deleted.\n",
+                        "schema": {
+                            "type": "number"
+                        }
+                    },
+                    "default": {
```
bradtaylor: 500 again?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    "200": {
+                        "description": "An array of configurations.",
+                        "schema": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/ConfigurationResponse"
+                            }
+                        }
+                    },
+                    "default": {
```
bradtaylor: 500 again?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                        "name": "user",
+                        "in": "query",
+                        "description": "User's email.",
+                        "required": true,
+                        "type": "string"
+                    },
+                    {
+                        "name": "roles",
+                        "in": "query",
+                        "description": "List of comma seperated roles -- i.e. Read,Write,Create,Redact,Manage or \"All\" or \"Nothing\". NO SPACES.",
```
bradtaylor: Is Put Nothing the same as Delete?

src/main/resources/swagger/agora.json
```diff
@@ -1,1353 +1,1442 @@
...
+                    "default": {
+                        "description": "Unexpected error",
+                        "schema": {
+                            "$ref": "#/definitions/Error"
+                        }
+                    }
+                }
+            },
+            "put": {
+                "summary": "Edit a permissions on namespace.",
```
bradtaylor: permission (singular). See also for method and configuration below

src/main/resources/swagger/agora.yaml
```diff
@@ -148,6 +149,38 @@ paths:
...
+          type: number
+      tags:
+        - Methods
+      responses:
+        200:
+          description: |
+            An integer showing how many methods were deleted.
+          schema:
+            type: number
+        default:
```
bradtaylor: Throughout, we have responses listed as "default", where we should probably be specifying their type (500, 400, etc).

Also, there are some cases where we have 400 responses, but #/definitions/Error only has 500. We should probably create some more error response definitions. Might as well do this now, while we're editing the file?

src/main/resources/swagger/agora.yaml
```diff
@@ -139,7 +140,7 @@ paths:
         - Methods
       responses:
         200:
-          description: An array of methods.
+          description: An single method.
           schema:
             type: array
```
bradtaylor: Getting a single method does not return an array. Might as well change this since we're here

src/main/resources/swagger/agora.yaml
```diff
@@ -909,17 +974,17 @@ definitions:
...
         description: Payload of method -- must be in WDL format
         default: |
-          task wc { command { cat ${File in_file} | wc -l } output { Int count = read_int(stdout()) }}
+          { "methodStoreMethod": { "methodNamespace": "YOUR_NAMESPACE", "methodName": "BWA", "methodVersion": 1}}
       entityType:
         type: string
-        description: Tyoe of the AgoraEntity -- Task or Workflow.
+        description: Type of the AgoraEntity -- Task or Workflow.
         default: Task
   MethodResponse:
```
bradtaylor: For both methods and configurations, payload and description are excluded by default in the return from the list method. (they are included when getting single entity). However, you CAN request them while doing a list by explicitly adding them to includedFields. Is there any way to make this clear?

src/main/resources/swagger/agora.yaml
```diff
@@ -285,6 +318,38 @@ paths:
...
+        200:
+          description: |
+            An integer showing how many configurations were deleted. Should be 1.
+          schema:
+            type: number
+        default:
+          description: Error upon redaction
+          schema:
+            $ref: '#/definitions/Error'
   /methods/{namespace}/permissions:
```
bradtaylor: All of the permissions methods are tagged as NamespacePermissions and Permissions respectively. So everything for /methods and /configurations end up in the same frame. Do you like this distinction? If it were me I might tag everything Methods or Configurations. So then things would be grouped by what kind of entity they refer to. Up to you

src/main/resources/swagger/agora.yaml
```diff
@@ -948,10 +1013,10 @@ definitions:
...
         description: Payload of method -- must be in WDL format
         default: |
-          task wc { command { cat ${File in_file} | wc -l } output { Int count = read_int(stdout()) }}
+          task wc {File in_file command { cat ${in_file} | wc -l } output { Int count = read_int(stdout()) }}
       entityType:
         type: string
-        description: Tyoe of the AgoraEntity -- Task or Workflow.
+        description: Type of the AgoraEntity -- Task or Workflow.
         default: Task
   AccessControl:
```
bradtaylor: For the description for roles, can we list the acceptable values? (Read, Write, Manage, Redact, All)

bradtaylor:
Some general comments, mostly related to things you didn't actually change. I figure it might be easy to fix these things up now.

Otherwise, thumbs up

PR #115, opened 2015-09-04
Added security definitions for google oauth2.

jacarey:


vivster7:
We should keep our swagger.json and swagger.yaml in sync. You can use http://editor.swagger.io/#/ to import a json file and generate a yaml file.

I find it easier to use their editor to edit the yaml, then generate the json.

jacarey:
yaml updated.

PR #118, opened 2015-09-11
Small bug fix to configuration validation

bradtaylor:
Rawls changed the configuration payload field "methodStoreMethod" to "methodRepoMethod"
We check for the existence of this field during validation, and we were checking by the old name.


vivster7:
hm, not sure why PR is closed, but looks good if you wanna merge it.

PR #119, opened 2015-09-11
Small bug fix to configuration validation

bradtaylor:
Rawls changed the configuration payload field "methodStoreMethod" to "methodRepoMethod"
We check for the existence of this field during validation, and we were checking by the old name.

Added better failure descriptions to our require statements

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.agora.server.business
 
 import cromwell.parser.BackendType
 import cromwell.parser.WdlParser.SyntaxError
-import org.broadinstitute.dsde.agora.server.exceptions.{AgoraEntityNotFoundException, NamespaceAuthorizationException}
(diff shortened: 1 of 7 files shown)


bradtaylor:
Do not merge tests failing

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -105,13 +105,18 @@ class AgoraBusiness {
...
         namespace.tasks.foreach { validateDockerImage }
       case AgoraEntityType.Configuration =>
-        val json = agoraEntity.payload.get.parseJson
-        val fields = json.asJsObject.getFields("methodStoreMethod")
-        require(fields.size == 1)
-        val subFields = fields(0).asJsObject.getFields("methodNamespace", "methodName", "methodVersion")
-        require(subFields(0).isInstanceOf[JsString])
-        require(subFields(1).isInstanceOf[JsString])
-        require(subFields(2).isInstanceOf[JsNumber])
+        try {
```
vivster7: why the try/ catch here? cant you just catch the IllegalArgumentException at the supervisor level. Then you can remove the ValidationException class altogether

PR #123, opened 2015-09-22
Removed supervisor error logging for tests.

jacarey:
Configurable with supervisor.logging which is true by default.
Also set whether or not to instrument Agora on application creation (could be in the config if we want to move it later)

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/Agora.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/Agora.scala
@@ -7,13 +7,13 @@ object ProductionAgora extends Agora {
     start()
 }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/Agora.scala
```diff
@@ -7,13 +7,13 @@ object ProductionAgora extends Agora {
     start()
 }
 
-class Agora extends LazyLogging with App {
+class Agora(instrument: Boolean = true) extends LazyLogging with App {
```
vivster7: use the agoraConfig value

vivster7:
:+1:

PR #126, opened 2015-10-01
public permissions

vivster7:
Support a ‘public’ user which can give permissions to all authenticated
users.


bradtaylor:
Thumbsup

PR #129, opened 2015-10-09
batch permissions POST

vivster7:


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/ServiceMessages.scala
```diff
@@ -19,6 +19,11 @@ object ServiceMessages {
                                         username: String,
                                         accessObject: AccessControl)
 
+  case class InsertBatchNamespacePermission(requestContext: RequestContext,
```
dshiga: Maybe just call this BatchNamespacePermission, since it can also be used to delete.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -30,19 +31,25 @@ abstract class AgoraService extends HttpService with RouteHelpers {
...
       parameterMap { (params) =>
-        val entity = AgoraEntity(Option(namespace))
+        val agoraEntity = AgoraEntity(Option(namespace))
 
+        // Accept batch POST // TODO: move to transactional support
+        entity(as[List[AccessControl]]) { (listOfAccessControl) =>
+          post { requestContext =>
+            completeBatchNamespacePermissionsPost(requestContext, agoraEntity, listOfAccessControl, username, permissionHandlerProps)
+          }
+        } ~
```
dshiga: Making sure I understand this - if it's a batch request we process it using the route above, and if it's a single request it falls through to the routes below?

vivster7: yup, it hits the entity directive if there is a body to their POST request -- I can refactor so they both fall under the post request below and remove duplication though

dshiga: Nah, just wanted to make sure we do fall through and haven't broken the single request route.

vivster7: decided not too -- harder to understand

PR #134, opened 2015-10-14
Add replicate mongo instances that can be added to the config. See sr…

jacarey:
…c/main/resources/reference.conf for example.


vivster7:
+1

PR #136, opened 2015-10-16
Merge develop changes to staging

jacarey:
This is to update the mongodb configuration to allow setting up replicate servers.


vivster7:
These don't need reviews, right? Since its already been reviewed for dev?

PR #139, opened 2015-10-19
added /api and versions /v1

vivster7:


src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -60,10 +60,11 @@ object AgoraConfig {
     case None => ""
     case x: Some[Int] => ":" + x.get
   }
-  private lazy val baseUrl = scheme + "://" + host + embeddedUrlPortStr + "/"
+  private lazy val baseUrl = scheme + "://" + host + embeddedUrlPortStr + "/" + "api" + "/" + version + "/"
   private lazy val scheme = config.as[Option[String]]("webservice.scheme").getOrElse("http")
   lazy val host = config.as[Option[String]]("webservice.host").getOrElse("localhost")
   lazy val port = config.as[Option[Int]]("webservice.port").getOrElse(8000)
+  lazy val version = config.as[Option[String]]("webservice.version").getOrElse("v1")
```
jacarey: How will this work when we have to support multiple version?  Maybe not in scope right now, but might be good to think about.

jacarey:
:+1:

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
vivster7: Yah, definitely merits a discussion -- we may just fork agora and spin up a new system for v2 as we deprecate v1. I'd imagine having two versions in the same codebase would be messy; however, I don't have much experience versioning.

PR #141, opened 2015-11-13
Poll official firecloud google groups for admins DSDEEPB-1388

bradtaylor:
Firecloud is containting their official list of admins in google groups for each environment.
If such a group is defined in config, poll for it's members, and synchronize our admin-user statuses with this list


abaumann:
looks good

PR #147, opened 2015-12-03
Support for arbitrary environments

abaumann:
rather than hardcoded environments, if it's local then it uses mock data and in mem mongo

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
@@ -15,44 +15,22 @@ object AgoraConfig {
 
   // Environments
   val LocalEnvironment = "local"
-  val DevEnvironment = "dev"
-  val StagingEnvironment = "staging"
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -15,45 +15,23 @@ object AgoraConfig {
...
-    usesEmbeddedMongo = false
-  }
-
-  // Prod
-  if (environment.equals(ProdEnvironment)) {
-    authenticationDirectives = OpenIdConnectDirectives
-    usesEmbeddedMongo = false
+  val environment = config.as[Option[String]]("environment")
+  environment match {
+    case Some(env) if env == "local" => {
```
bradtaylor: use localEnvironment from above, instead of hardcoded string

bradtaylor:
1 comment, otherwise thumbsup

abaumann:
addressed and rebased

PR #148, opened 2015-12-08
Require that configs reference workflows (not tasks)

dshiga:
We want to disallow inserting a configuration that references a task -- it must reference a workflow.

The only real changes are in AgoraBusiness. (Also added a test in AgoraConfigurationsSpec -- the rest of the changes are due to renaming test entities).

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -27,11 +27,7 @@ class AgoraBusiness {
     val entityToInsert = agoraEntity.entityType.get match {
       case AgoraEntityType.Configuration =>
         val method = resolveMethodRef(agoraEntity.payload.get)
-        if (!AgoraEntityPermissionsClient.getEntityPermission(method, username).canRead &&
-            !AgoraEntityPermissionsClient.getEntityPermission(method, "public").canRead) {
-          throw new AgoraEntityNotFoundException(method)
-        }
-
+        validateMethodRef(method, username)
         agoraEntity.addMethodId(method.id.get.toHexString)
       case _ => agoraEntity
     }
@@ -128,6 +124,18 @@ class AgoraBusiness {
     }
   }
 
+  private def validateMethodRef(method: AgoraEntity, username: String) = {
+    if (!AgoraEntityPermissionsClient.getEntityPermission(method, username).canRead &&
+      !AgoraEntityPermissionsClient.getEntityPermission(method, "public").canRead) {
+      throw new AgoraEntityNotFoundException(method)
+    }
+    if (method.entityType.get == AgoraEntityType.Task) {
+      val methodIdentifier = "%s/%s/%s".format(method.namespace.get, method.name.get, method.snapshotId.get)
+      val message = "Configurations can only reference workflows, not tasks. Attempted to reference task %s".format(methodIdentifier)
+      throw new ValidationException(message)
+    }
+  }
+  
   private def validateDockerImage(task: Task) = {
     if (task.runtimeAttributes.docker.isDefined) {
       val dockerImageReference = parseDockerString(task.runtimeAttributes.docker.get)
--- a/src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
+++ b/src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
@@ -26,6 +26,8 @@ object AgoraTestData {
   val synopsis3 = Option("This is a test configuration")
   val documentation1 = Option("This is the documentation")
   val documentation2 = Option("This is documentation for another method")
+  val taskNamespace2 = Option("taskNamespace2")
+  val taskName2 = Option("taskName2")
 
   val badNamespace = Option("    ")
   val badName = Option("   ")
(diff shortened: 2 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -128,6 +124,18 @@ class AgoraBusiness {
...
   }
 
+  private def validateMethodRef(method: AgoraEntity, username: String) = {
+    if (!AgoraEntityPermissionsClient.getEntityPermission(method, username).canRead &&
+      !AgoraEntityPermissionsClient.getEntityPermission(method, "public").canRead) {
+      throw new AgoraEntityNotFoundException(method)
+    }
+    if (method.entityType.get == AgoraEntityType.Task) {
+      val methodIdentifier = "%s/%s/%s".format(method.namespace.get, method.name.get, method.snapshotId.get)
+      val message = "Configurations can only reference workflows, not tasks. Attempted to reference task %s".format(methodIdentifier)
```
bradtaylor, line 134: Can use scala string interpolation rather than .format:
s"Configurations can only reference workflows, not tasks. Attempted to reference task ${methodIdentifier}"

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -128,6 +124,18 @@ class AgoraBusiness {
     }
   }
 
+  private def validateMethodRef(method: AgoraEntity, username: String) = {
+    if (!AgoraEntityPermissionsClient.getEntityPermission(method, username).canRead &&
+      !AgoraEntityPermissionsClient.getEntityPermission(method, "public").canRead) {
+      throw new AgoraEntityNotFoundException(method)
```
bradtaylor, line 130: Do we really want to throw an entity not found exception? The config exists, but it's referenced method doesn't. So is that a ValidationException?  How is this exception being handled?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
dshiga, line 134: Good call

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -399,7 +425,17 @@ object AgoraTestData {
     payload = payload1,
     entityType = Option(AgoraEntityType.Workflow))
 
-  val testEntity7 = AgoraEntity(namespace = namespace1,
+  val testWorkflow7 = new AgoraEntity(
```
bradtaylor, line 428: We don't need the 'new' here, right?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -26,6 +26,8 @@ object AgoraTestData {
   val synopsis3 = Option("This is a test configuration")
   val documentation1 = Option("This is the documentation")
   val documentation2 = Option("This is documentation for another method")
+  val taskNamespace2 = Option("taskNamespace2")
```
bradtaylor, line 29: Why '2'? This is the first task-specific name I see.
Should we just add a name5?  It feels like our convention here is getting inconsistent

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
dshiga, line 130: We are intentionally being ambiguous here because we don't want the user to be able to tell if a method truly doesn't exist vs exists but not readable to them. So we return a 404 in either case.

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -407,7 +443,17 @@ object AgoraTestData {
     payload = payload2,
     entityType = Option(AgoraEntityType.Task))
 
-  val testEntityToBeRedacted = AgoraEntity(namespace = namespace1,
+  val testTask2 = new AgoraEntity(
```
bradtaylor, line 446: new?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -399,7 +425,17 @@ object AgoraTestData {
...
+    name = name1,
+    synopsis = synopsis1,
+    documentation = documentation1,
+    owner = owner1,
+    payload = payload1,
+    entityType = Option(AgoraEntityType.Workflow)
+  )
+
+  val testTask1 = AgoraEntity(namespace = namespace1,
     name = name2,
```
bradtaylor, line 439: Should there be a taskName1?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -639,22 +689,4 @@ object AgoraTestData {
     payload = payloadWithInvalidPersonalDockerTagNameInWdl,
     entityType = Option(AgoraEntityType.Task)
   )
-
-  def populateDatabase(agoraBusiness: AgoraBusiness) = {
```
bradtaylor, line 643: is this function not used?

dshiga, line 643: Apparently not. I did a search in IntelliJ and found no references. All the tests pass happily without it.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -63,49 +71,55 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
     }
   }
 
+  "Agora" should "not allow you to add a configuration that references a task" in {
+    Post(ApiUtil.Configurations.withLeadingVersion, testBadConfigReferencesTask) ~>
+      configurationsService.postRoute ~> check {
+      assert(status === BadRequest)
```
bradtaylor, line 77: What is handleError and does this line need it?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
bradtaylor, line 130: :+1:

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
bradtaylor, line 643: I think it was left over from before we implemented the beforeAll / afterAll db management stuff. Happy to see it killed

bradtaylor:
A couple very minor questions/comments. Otherwise  :+1:

davidangb:
abandoning this PR.

PR #149, opened 2016-01-05
Fix build error due to conflicting versions of guava library

dshiga:
The Agora build was broken because we were pulling in conflicting versions of the guava library (required by Google libraries through their chains of dependencies).

This wasn't happening before but some of those libraries must have changed to produce the conflict.

Fixed by adding a rule to build.sbt.


abaumann:
looks good to me

PR #150, opened 2016-01-06
Standardize how we docker

dshiga:
This brings Agora in line with how the rest of the teams handle dockerizing. It also allows us to pass in some options to java when we start Agora (some certs need to be passed in to support talking to mysql via ssl).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,16 @@
 # Production build/deploy/start of DSDE Methods Repository Webservice
 
-# Basis image- Lukas K. recommended
-FROM debian:jessie
-MAINTAINER DSDE <dsde-engineering@broadinstitute.org>
+# http://github.com/broadinstitute/scala-baseimage
+FROM broadinstitute/scala-baseimage
 
-
-# Install necessary packages including java 8 jre and sbt and clean up apt caches
-RUN echo "deb http://ppa.launchpad.net/webupd8team/java/ubuntu trusty main" >> /etc/apt/sources.list.d/webupd8team-java.list && \
-    echo "deb-src http://ppa.launchpad.net/webupd8team/java/ubuntu trusty main" >> /etc/apt/sources.list.d/webupd8team-java.list && \
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys EEA14886 && \
-    echo debconf shared/accepted-oracle-license-v1-1 select true | /usr/bin/debconf-set-selections && \
-    echo debconf shared/accepted-oracle-license-v1-1 seen true | /usr/bin/debconf-set-selections
-
-RUN apt-get update && \
-    apt-get --no-install-recommends install -y --force-yes oracle-java8-installer &&\
-    apt-get clean autoclean && \
-    apt-get autoremove -y && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,29 +1,16 @@
...
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys EEA14886 && \
-    echo debconf shared/accepted-oracle-license-v1-1 select true | /usr/bin/debconf-set-selections && \
-    echo debconf shared/accepted-oracle-license-v1-1 seen true | /usr/bin/debconf-set-selections
-
-RUN apt-get update && \
-    apt-get --no-install-recommends install -y --force-yes oracle-java8-installer &&\
-    apt-get clean autoclean && \
-    apt-get autoremove -y && \
-    rm -rf /var/lib/{apt,dpkg,cache,log}/ /var/cache/oracle-jdk8-installer 
-
```
dshiga, line 20: No need to run all the above commands anymore because the scala base image provides all of this.

Dockerfile
```diff
@@ -1,29 +1,16 @@
...
-# Expose the port used by the webservice
+# Expose the port used by Agora webservice
 EXPOSE 8000
 
-WORKDIR /usr/agora
-
-COPY agora-0.1-SNAPSHOT.jar /usr/agora/
+# Install Agora
+ADD . /agora
+RUN ["/bin/bash", "-c", "/agora/docker/install.sh /agora"]
```
dshiga, line 11: More standard to call an install script here which builds the jar and copies it where needed.

Dockerfile
```diff
@@ -1,29 +1,16 @@
...
 
-# Start the webservice with default parameters do not use default config
-ENTRYPOINT ["java", "-Djava.library.path=./native", "-Dconfig.file=/etc/agora.conf", "-javaagent:agora-0.1-SNAPSHOT.jar", "-jar", "agora-0.1-SNAPSHOT.jar"]
+# Add Agora as a service (it will start when the container starts)
+RUN mkdir /etc/service/agora && \
+    mkdir /var/log/agora && \
+    cp /agora/docker/run.sh /etc/service/agora/run
```
dshiga, line 16: Again, following what the other teams are doing, we install agora as a "runit" service, which will start on system startup and restart if it does. More here:
https://www.debian-administration.org/article/697/Using_runit_for_maintaining_services

docker/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+set -e
+java $JAVA_OPTS -Djava.library.path=./native -Dconfig.file=/etc/agora.conf -javaagent:/agora/agora.jar -jar /agora/agora.jar
```
dshiga, line 4: $JAVA_OPTS can/will be passed to the container when starting it up.

kcibul:
looks good dave -- there aren't any automated tests at the moment that would catch if this docker image is borked.  @ndbolliger -- this would be a great thing to know that some automated integration tests would catch (ie if the docker build was bad so the service didn't stand up, that the tests would fail)

dshiga:
Thanks @kcibul. It would definitely be nice to have automated docker tests - we've been bitten in the past by changes that broke our docker build. In this case I tested docker manually on my machine, but it's still not a 100% guarantee because the environment is different than our real deployments.

PR #151, opened 2016-01-07
Changes to the proxy settings to support LDAP Authorization.

lukwam:


src/main/config/proxy-compose.yaml.ctmpl
```diff
@@ -7,15 +7,11 @@ proxy:
   {{.Data.proxy_ports}}
   {{.Data.proxy_volumes}}
   environment:
-    AUTH_TYPE: AuthType None
-    AUTH_TYPE2: AuthType oauth20
     CALLBACK_URI: {{.Data.env_callback_uri}}
-    LOG_LEVEL: warn
-    OIDC_CLAIM: Require all granted
```
bradtaylor, line 14: We are currently getting the user's email address from the request header via the "OIDC_CLAIM_email" value. That email becomes the primary thing we rely on to keep track of our users. Is that going away?

lukwam, line 14: The change from email to sub means that the REMOTE_USER environment variable will now have the numeric value of the google userid instead of the email.

The OIDC_CLAIM_email will ALWAYS have the email address stored in it.

bradtaylor:
:+1:

PR #154, opened 2016-01-11
Use least powerful scope needed to query admin group

dshiga:
We only need the member read only scope. We were using a broader scope providing read/write access to the group.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/GoogleApiUtils.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/GoogleApiUtils.scala
@@ -17,7 +17,7 @@ object GoogleApiUtils {
   val emailAddress = AgoraConfig.gcsServiceAccountEmail
   val JSON_FACTORY = JacksonFactory.getDefaultInstance
   val httpTransport = GoogleNetHttpTransport.newTrustedTransport
-  val directoryScopes = Collections.singleton(DirectoryScopes.ADMIN_DIRECTORY_GROUP)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/GoogleApiUtils.scala
```diff
@@ -17,7 +17,7 @@ object GoogleApiUtils {
   val emailAddress = AgoraConfig.gcsServiceAccountEmail
   val JSON_FACTORY = JacksonFactory.getDefaultInstance
   val httpTransport = GoogleNetHttpTransport.newTrustedTransport
-  val directoryScopes = Collections.singleton(DirectoryScopes.ADMIN_DIRECTORY_GROUP)
+  val directoryScopes = Collections.singleton(DirectoryScopes.ADMIN_DIRECTORY_GROUP_MEMBER_READONLY)
```
bradtaylor, line 20: Did we want to move this into config? Maybe we make a non-1/20 background story for that, as doing it would require a concomitant vault change

bradtaylor:
Harcode is fine for now. If this works locally then :+1:

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/GoogleApiUtils.scala
dshiga, line 20: Good thought. But I don't think we should ever need a different scope unless we change the code to do something more than read. So we wouldn't really save ourselves any trouble by making this configurable.

PR #156, opened 2016-01-19
Disabled docker image validation.

gbggrant:
Disabled unit tests that check docker image validation

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.agora.server.business
 
 import cromwell.parser.BackendType
 import cromwell.parser.WdlParser.SyntaxError
-import org.broadinstitute.dsde.agora.server.exceptions.{ValidationException, AgoraException, AgoraEntityNotFoundException, NamespaceAuthorizationException}
+import org.broadinstitute.dsde.agora.server.exceptions.{ValidationException, AgoraEntityNotFoundException, NamespaceAuthorizationException}
 import org.broadinstitute.dsde.agora.server.webservice.util.{DockerImageReference, DockerHubClient}
 
 import cromwell.binding._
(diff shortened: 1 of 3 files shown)


src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessIntegrationSpec.scala
```diff
@@ -22,14 +22,14 @@ class AgoraBusinessIntegrationSpec extends FlatSpec with BeforeAndAfterAll with
     assert(agoraBusiness.findSingle(entity, Seq(entity.entityType.get), mockAutheticatedOwner.get) === entity)
   }
 
-  "Agora" should "be unable to store a task configuration with an invalid docker image (invalid/non-existent repo name)" in {
+  ignore should "be unable to store a task configuration with an invalid docker image (invalid/non-existent repo name)" in {
```
bradtaylor: This isn't your thing (I think this is fine), but are we running this test twice? Why?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -129,12 +127,16 @@ class AgoraBusiness {
   }
 
   private def validateDockerImage(task: Task) = {
-    if (task.runtimeAttributes.docker.isDefined) {
-      val dockerImageReference = parseDockerString(task.runtimeAttributes.docker.get)
-      if (dockerImageReference.isDefined) {
-        DockerHubClient.doesDockerImageExist(dockerImageReference.get)
-      }
-    }
+    // Per DSDEEPB-2525, in the interests of expediency, we are disabling docker image validation as we do not support validation of private docker images
```
bradtaylor, line 130: I think this is fine, but for my edification is there a reason to turn this into a trivial function rather than commenting out the calls in AgoraBusiness.insert?

bradtaylor:
:+1: from me

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
gbggrant, line 130: I left it here as a place holder since it seems we'll be adding the functionality back in soon.

PR #159, opened 2016-01-27
remove user from ACL when given no access

MatthewBemis:
For [DSDEEPB-1970](https://broadinstitute.atlassian.net/browse/DSDEEPB-1970)

I wasn't able to stand up a local instance of Agora to test this out, so it's tested solely through modifications to existing unit tests, which I've reverted for this PR. If you'd like, I can write a unit test explicitly for this, but since this change wasn't any actual new functionality I wasn't sure if that was appropriate.


vivster7:
:+1:

PR #163, opened 2016-03-08
Update Flyway repo URL to fix build

dmohs:


abaumann:
hmm interesting, any explanation why that was only a problem in the docker build?

dmohs:
I'd say it's because a docker build is generally cleaner than a local build.

abaumann:
:+1:

PR #165, opened 2016-09-02
GAWB-773 added agora status endpoint to check connection to sql and mongo dbs

skwalker:
  - Acceptance criteria exists and is met
  - Note any changes to implementation from the description
  - Add notes on what you've tested
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
  - Rest of team may comment on PR at will
  - **LR assigns to submitter** for feedback fixes
  - Submitter rebases to develop again if necessary
  - Submitter makes further commits. DO NOT SQUASH
  - Submitter updates documentation as needed
  - Submitter **reassigns to LR** for further feedback
- [x] **TL** sign off
- [ ] **LR** sign off


src/test/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceSpec.scala
```diff
@@ -28,8 +28,15 @@ class ApiServiceSpec extends FlatSpec with Directives with ScalatestRouteTest wi
     def actorRefFactory = system
   }
 
+
+  abstract class StatusService extends AgoraService {
```
skwalker: wasn't really sure where this should go / if there's a better way to do this

dvoet:
Overall feedback is that this does not fit into the overall agora architecture. You should probably be using the completeWithPerRequest function in AgoraService. Try following the example of org.broadinstitute.dsde.agora.server.webservice.handlers.AddHandler.

Also, the current code executes the queries to mongo and mysql serially and we would only see the mysql error if mongo fails. We should do these in parallel and report both errors independently.

skwalker:
@dvoet responded to your comment

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraStatus.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.broadinstitute.dsde.agora.server.dataaccess.mongo.AgoraMongoClient._
+import org.broadinstitute.dsde.agora.server.dataaccess.permissions.AgoraEntityPermissionsClient._
+
+class AgoraStatus {
+  def status(): (Boolean, String) = {
+    val (up, message) = getMongoDBStatus
+    try {
+      // sql up, mongo may be up or down
+      sqlDBStatus
+      return (up, s"""{"status": "down", "mongo error": "$message" }""")
```
dvoet: returns are bad in scala. Anonymous functions are embedded all over the place so it can be hard to tell what you are returning from. In this case you just don't need it at all as the last value is the return value.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraStatus.scala
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.dsde.agora.server.dataaccess
+
+import org.broadinstitute.dsde.agora.server.dataaccess.mongo.AgoraMongoClient._
+import org.broadinstitute.dsde.agora.server.dataaccess.permissions.AgoraEntityPermissionsClient._
+
+class AgoraStatus {
+  def status(): (Boolean, String) = {
+    val (up, message) = getMongoDBStatus
```
dvoet: the pattern for getMongoDBStatus is different than sqlDBStatus, the former does not throw and exception but the latter does.

I would change the signature to something like

checkMongoConnection(): Try[Unit]
checkSqlConnection(): Try[Unit]

I don't really like Unit but the return value on success does not really matter. When there is an exception you can extract the message from the failure. Then you can do something like

val mongo = checkMongoConnection()
val sql = checkSqlConnection()

val errors = Set(("mongo error", mongo), ("sql error", sql)).collect { case (label, Failure(t)) => s""""$label": "$t.message""""
if (errors.isEmpty) {
  """{"status": "up"}"""
} else {
  s"""{ "status": "down", ${errors.mkString(", ") }"""
}

That also centralizes all the json ick

skwalker: This looks much better, but if the status method just returns the string, then wouldn't it always be a 200 OK even when status is down?

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/AgoraDBStatus.scala
```diff
@@ -0,0 +1,19 @@
...
+
+class AgoraDBStatus {
+  def status(): AgoraStatus = {
+    val mongo = getMongoDBStatus
+    val sql = sqlDBStatus
+
+    val errors = Seq(mongo, sql).collect {
+      case (Failure(t)) => t.getMessage
+    }
+    if (errors.isEmpty) AgoraStatus(true, Seq()) else AgoraStatus(false, errors)
```
dvoet: minor nitpick: no if required:
AgoraStatus(errors.isEmpty, errors)

dvoet (approved):

ansingh7115:
👍

PR #169, opened 2016-10-27
Make swagger work and add support for running locally under docker

rushtong:
This addresses [GAWB-1190](https://broadinstitute.atlassian.net/browse/GAWB-1190)
- Updates swagger so oauth works locally
- Add sbt revolver so a local instance can be stood up similar to our standard Rawls/Orchestration/UI stack. This does not integrate with that stack, but sets us up for doing so.
- Look at https://github.com/broadinstitute/firecloud-develop/pull/166 for the configs that need to be rendered.
- When this is migrated to other environments, we need to add a web service account google client id to the vault secrets per environment. 

---
  - Acceptance criteria exists and is met
  - Note any changes to implementation from the description
  - Add notes on what you've tested
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
  - Rest of team may comment on PR at will
  - **LR assigns to submitter** for feedback fixes
  - Submitter rebases to develop again if necessary
  - Submitter makes further commits. DO NOT SQUASH
  - Submitter updates documentation as needed
  - Submitter **reassigns to LR** for further feedback
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email

### DIFF
--- a/docker/run-local.sh
+++ b/docker/run-local.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+cd /agora
+sbt "~reStart --- $JAVA_OPTS"
--- a/docker/run-local.sh
+++ b/docker/run-local.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
-
-set -e
+set -euox pipefail
+IFS=$'\n\t'
 
 cd /agora
-sbt "~reStart --- $JAVA_OPTS"
+sbt ~reStart
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -25,6 +25,9 @@ swagger {
   licenseUrl = "http://opensource.org/licenses/BSD-3-Clause"
   swaggerVersion = "1.3"
   termsOfServiceUrl = "http://www.github.com/broadinstitute/agora"
+  clientId = "your-client-id"
+  realm = "your-realm"
+  appName = "your-app-name"
 }
 
 akka {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -26,8 +26,6 @@ swagger {
   swaggerVersion = "1.3"
   termsOfServiceUrl = "http://www.github.com/broadinstitute/agora"
   clientId = "your-client-id"
-  realm = "your-realm"
-  appName = "your-app-name"
 }
 
 akka {
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
@@ -83,8 +83,8 @@ object AgoraConfig {
     lazy val baseUrl = swagger.getString("baseUrl")
     lazy val apiDocs = swagger.getString("apiDocs")
     lazy val clientId = swagger.as[Option[String]]("clientId").getOrElse("clientId")
-    lazy val realm = swagger.as[Option[String]]("realm").getOrElse(gcsProjectId)
-    lazy val appName = swagger.as[Option[String]]("appName").getOrElse(gcsProjectId)
+    lazy val realm = gcsProjectId
+    lazy val appName = gcsProjectId
   }
 
   def urlFromType(entityType: Option[EntityType]): String = {
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
@@ -82,6 +82,9 @@ object AgoraConfig {
     lazy val licenseUrl = swagger.getString("licenseUrl")
     lazy val baseUrl = swagger.getString("baseUrl")
     lazy val apiDocs = swagger.getString("apiDocs")
+    lazy val clientId = swagger.as[Option[String]]("clientId").getOrElse("clientId")
+    lazy val realm = swagger.as[Option[String]]("realm").getOrElse(gcsProjectId)
+    lazy val appName = swagger.as[Option[String]]("appName").getOrElse(gcsProjectId)
   }
 
   def urlFromType(entityType: Option[EntityType]): String = {
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
@@ -65,25 +65,15 @@ class ApiServiceActor extends HttpServiceActor with LazyLogging {
       pathSingleSlash {
         withResourceFileContents("swagger/index.html") { indexHtml =>
           complete {
-            val swaggerOptions =
-              """
-                |        validatorUrl: null,
-                |        apisSorter: "alpha",
-                |        operationsSorter: "alpha",
-              """.stripMargin
             HttpEntity(ContentType(MediaTypes.`text/html`),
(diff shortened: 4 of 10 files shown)


docker/run-local.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+cd /agora
+sbt "~reStart --- $JAVA_OPTS"
```
davidangb: potential follow-on story, or part of this one if you want to tackle it; it looks small: see https://github.com/broadinstitute/rawls/pull/527#r78079727. David Mohs figured out how to simplify this awkward restart syntax.

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -82,6 +82,9 @@ object AgoraConfig {
     lazy val licenseUrl = swagger.getString("licenseUrl")
     lazy val baseUrl = swagger.getString("baseUrl")
     lazy val apiDocs = swagger.getString("apiDocs")
+    lazy val clientId = swagger.as[Option[String]]("clientId").getOrElse("clientId")
+    lazy val realm = swagger.as[Option[String]]("realm").getOrElse(gcsProjectId)
+    lazy val appName = swagger.as[Option[String]]("appName").getOrElse(gcsProjectId)
```
davidangb: since these are populated in reference.conf, do you need the getOrElse?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
```diff
@@ -55,11 +57,32 @@ class ApiServiceActor extends HttpServiceActor with LazyLogging {
...
+              """.stripMargin
+            HttpEntity(ContentType(MediaTypes.`text/html`),
+              indexHtml
+                .replace("your-client-id", AgoraConfig.SwaggerConfig.clientId)
+                .replace("your-realms", AgoraConfig.SwaggerConfig.realm)
+                .replace("your-app-name", AgoraConfig.SwaggerConfig.appName)
+                .replace("scopeSeparator: \",\"", "scopeSeparator: \" \"")
+                .replace("jsonEditor: false,", "jsonEditor: false," + swaggerOptions)
+                .replace("url = \"http://petstore.swagger.io/v2/swagger.json\";",
+                  "url = '/api-docs';")
```
davidangb: should this be `url=agora.yaml`, not `url=api-docs`?

davidangb (commented):
I can start and run this, but swagger won't authenticate me. I was able to troubleshoot it down to the point that sbt reStart isn't reading from /etc/agora.conf - even though I do have `-Dconfig.file=/etc/agora.conf` in JAVA_OPTS, and /etc/agora.conf exists and has the right values. Don't know WHY it's not working yet, but figured I'd get this review out.

Update: after restarting and repulling everything, I have swagger working and reading from /etc/agora.conf. I can read permissions from mysql, but I get errors anytime I try to use mongo, such as listing methods. Should that be working?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
```diff
@@ -55,11 +57,32 @@ class ApiServiceActor extends HttpServiceActor with LazyLogging {
...
+                |        apisSorter: "alpha",
+                |        operationsSorter: "alpha",
+              """.stripMargin
+            HttpEntity(ContentType(MediaTypes.`text/html`),
+              indexHtml
+                .replace("your-client-id", AgoraConfig.SwaggerConfig.clientId)
+                .replace("your-realms", AgoraConfig.SwaggerConfig.realm)
+                .replace("your-app-name", AgoraConfig.SwaggerConfig.appName)
+                .replace("scopeSeparator: \",\"", "scopeSeparator: \" \"")
+                .replace("jsonEditor: false,", "jsonEditor: false," + swaggerOptions)
```
davidangb: also, this line isn't going to work - `jsonEditor` was added in a later version of swagger-ui. You'll have to either upgrade swagger-ui, or find something else to hang the replacement off of. Or ... just edit index.html directly, since it seems agora copied it in.

rushtong:
~~Mongo requires edits to /etc/hosts such that the instance names are mapped to their external IP addresses. I'll update the readme with that info.~~

Update: this isn't true. Will look into the mongo thing some more.

Update: This was partially true. We needed to add an extra_hosts line to the compose so the mongo library knows how to map configured mongo instance names to their external IP addresses. This is now in place in the corresponding firecloud-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/166

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
rushtong: I missed that agora pulled in a copy. I also think we should upgrade swagger here.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiServiceActor.scala
rushtong: The api-docs works for me locally. There is an endpoint in agora that does this so I think both cases will work.

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
rushtong: This could be even simpler since the same value is in `gcsProjectId`, but I'll test to see if the value in reference.conf is available to AgoraConfig.

docker/run-local.sh
rushtong: Nice! I was referencing our existing code and also this new how-to doc: https://broadinstitute.atlassian.net/wiki/display/GAWB/How+to+pass+Java+options+to+sbt+~reStart

davidangb:
I have not tested. Code looks good, with one exception: even though sbt clean/compile/assembly works, sbt reStart does not. I think the `Revolver` code in build.sbt is for 0.7.2, not 0.8.0.

rushtong:
Will look again at sbt reStart - it's working for me locally. I did have to change the syntax of a couple things based on v0.8.0.

src/main/resources/reference.conf
```diff
@@ -25,6 +25,7 @@ swagger {
   licenseUrl = "http://opensource.org/licenses/BSD-3-Clause"
   swaggerVersion = "1.3"
   termsOfServiceUrl = "http://www.github.com/broadinstitute/agora"
+  clientId = "your-client-id"
```
skwalker, line 28: do we still need reference.conf?

skwalker:
Everything seems to be working for me -- this is awesome Greg, thank you! 👍 from me

ahaessly:
i was able to get agora running locally via docker and use the swagger interface to retrieve methods and configurations :+1:

davidangb:
👍 latest changes work great for me. I have two issues for a follow-on story, not for this PR:
1) `sv stop agora` doesn't work (we have the same problem in orch)
2) I'd like to see a full-stack docker-compose with UI/orch/agora, with or without rawls as well

But good to merge!

davidangb (approved):

rushtong:
Follow on stories created and linked to the the ticket for this PR.

PR #170, opened 2016-11-02
GAWB-1161: Agora show owners of methods and method configs

skwalker:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -82,13 +80,9 @@ class AgoraBusiness {
...
                  username: String): AgoraEntity = {
     val foundEntity = AgoraDao.createAgoraDao(entityTypes).findSingle(namespace, name, snapshotId)
-
-    if (!AgoraEntityPermissionsClient.getEntityPermission(foundEntity, username).canRead &&
-        !AgoraEntityPermissionsClient.getEntityPermission(foundEntity, "public").canRead) {
-      throw new AgoraEntityNotFoundException(foundEntity)
-    }
-
-    foundEntity.addUrl().removeIds()
+    val entities = AgoraEntityPermissionsClient.filterEntityByRead(Seq(foundEntity), username)
```
skwalker, line 83: this makes checking for read permissions way faster since it checks username and public in the same sql query

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -126,6 +126,25 @@ trait PermissionsClient {
 
   }
 
+  def listOwners(agoraEntity: AgoraEntity): Seq[String] = {
+    val permissionsQuery = for {
+      entity <- entities if entity.alias === alias(agoraEntity)
+      _permissions <- permissions if _permissions.entityID === entity.id // && _permissions.roles===Manage
```
skwalker: I'm not sure if I should be checking to see if you have manage/owner permissions within the sql query itself instead of filtering the future/result like what I'm currently doing

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -76,6 +76,7 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
       if (entity.synopsis.nonEmpty) map += ("synopsis" -> JsString(entity.synopsis.get))
       if (entity.documentation.nonEmpty) map += ("documentation" -> JsString(entity.documentation.get))
       if (entity.owner.nonEmpty) map += ("owner" -> JsString(entity.owner.get))
+      if (entity.managers.size > 0) map += ("owners" -> JsArray(entity.managers.map(JsString(_)).toVector))
```
davidangb: did you mean "managers", not "owners"?

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessIntegrationSpec.scala
```diff
@@ -18,46 +18,47 @@ class AgoraBusinessIntegrationSpec extends FlatSpec with BeforeAndAfterAll with
   }
 
   "Agora" should "be able to store a task configuration with a valid (extant) official docker image" in {
-    val entity = agoraBusiness.insert(testAgoraEntityWithValidOfficialDockerImageInWdl, mockAutheticatedOwner.get)
-    assert(agoraBusiness.findSingle(entity, Seq(entity.entityType.get), mockAutheticatedOwner.get) === entity)
+    val entity = agoraBusiness.insert(testAgoraEntityWithValidOfficialDockerImageInWdl, mockAuthenticatedOwner.get)
+    val actual = agoraBusiness.findSingle(entity, Seq(entity.entityType.get), mockAuthenticatedOwner.get)
+    assert(actual === entity.addManagers(Seq(mockAuthenticatedOwner.get)))
```
davidangb: for code style, I'd prefer to do all the manipulation in one place, and keep the assert simple:

```
val entity = agoraBusiness.insert(testAgoraEntityWithValidOfficialDockerImageInWdl, mockAuthenticatedOwner.get)
  .addManagers(Seq(mockAuthenticatedOwner.get))
val actual = agoraBusiness.findSingle(entity, Seq(entity.entityType.get), mockAuthenticatedOwner.get)
assert(actual === entity)
```

(untested code, but you get the gist)

also adjust the test at line 42

README.md
```diff
@@ -22,7 +22,7 @@ $ sbt package
 You can additionally run unit test through:
 
 ```
-$ sbt unitTest
+$ sbt test -Dconfig.file=src/main/resources/reference.conf
```
davidangb: this should be `src/test` ?

It also shouldn't be necessary, based on what's in build.sbt. That may take some investigation to see why your environment needs it.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -82,13 +80,9 @@ class AgoraBusiness {
...
+    if (entities.size == 0) throw new AgoraEntityNotFoundException(foundEntity)
+    else entities.head.addUrl().removeIds().addManagers(AgoraEntityPermissionsClient.listOwners(foundEntity))
```
davidangb: I'd want to check the weird corner case where size > 1 here. Your current code won't throw an error on this, and I think we should, because it's totally unexpected. You could check `entities.size == 1`, or you could get all Scala-happy and pattern-match it, looking for a sequence of one element:

```
    entities match {
      case Seq(ae:AgoraEntity) => ae.addUrl().removeIds().addManagers(AgoraEntityPermissionsClient.listOwners(foundEntity))
      case _ => throw new AgoraEntityNotFoundException(foundEntity)
    }
```

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
davidangb: yeah, you should absolutely leverage the database to do this. Looking at other agora code such as `AgoraBusiness.filterEntityByRead()` ... please don't string together a list of `permission.roles === 1 || permission.roles === 3 || permission.roles === 5 ...`. Look at Slick's `inSetBind` instead.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -126,6 +126,25 @@ trait PermissionsClient {
...
+    val permissionsFuture = db.run(permissionsQuery.result)
+
+    val permissionsRole = permissionsFuture.map { objects: Seq[(String, Int)] =>
+      objects.filter { obj =>
+        AgoraPermissions(obj._2).canManage
+      }.map { obj => obj._1 }
+    } recover {
+      case ex: Throwable => throw new PermissionNotFoundException(s"Could not find owner? ", ex)
+    }
+    Await.result(permissionsRole, timeout)
```
davidangb: Boo. We should avoid `Await.result`, which blocks a thread, and return a Future instead, only map that Future when we really need to.

BUT ... looking at other Agora code, `Await.result` is all over the place, and I'm sure you modeled your code after that. If changing it in this one location is difficult, go ahead and keep it here, and let's file a story to clean it up for real.

davidangb (changes requested):
see comments inline!

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -126,6 +126,20 @@ trait PermissionsClient {
 
   }
 
+  def listOwners(agoraEntity: AgoraEntity): Seq[String] = {
+    // you're a manager if your role & (1 << 4) != 0 (and I think we're in the range from 0-32) so basically 16-31 inclusive
+    val permissionsQuery = for {
+      entity <- entities if entity.alias === alias(agoraEntity)
+      _permissions <- permissions if _permissions.entityID === entity.id && (_permissions.roles inSetBind List.range(16, 32))
+      user <- users if user.id === _permissions.userID
+    } yield user.email
```
davidangb, line 134: Instead of hardcoding numbers here, please use the variables in `AgoraPermissions`. Perhaps you can just check for `>= AgoraPermissions.Manage`? There's no need right now to check less-than 32, is there? That would only be relevant if we added a new permission (and that permission was less powerful than `Manage`.

I know I pointed you at `inSetBind` earlier, and maybe that led you astray - I didn't know how the permissions were actually stored, sorry!

rushtong: +1 to what David said.

I find the `_permissions`  confusing to read. I think `permission <- permissions` is more legible.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -126,6 +126,20 @@ trait PermissionsClient {
...
+    try {
+      Await.result(db.run(permissionsQuery.result), timeout)
```
rushtong, line 136: There are a lot of Awaits here. I think we need a tech debt story to refactor these into futures.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -3,11 +3,11 @@ package org.broadinstitute.dsde.agora.server.model
 
 import org.broadinstitute.dsde.agora.server.exceptions.AgoraException
 import org.broadinstitute.dsde.agora.server.webservice.util.AgoraOpenAMClient.UserInfoResponse
-import org.broadinstitute.dsde.agora.server.dataaccess.permissions.{AgoraPermissions, AccessControl}
+import org.broadinstitute.dsde.agora.server.dataaccess.permissions.{AccessControl, AgoraPermissions}
```
rushtong: This isn't actually a change - it will be a cleaner commit to revert it.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -76,6 +76,7 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
       if (entity.synopsis.nonEmpty) map += ("synopsis" -> JsString(entity.synopsis.get))
       if (entity.documentation.nonEmpty) map += ("documentation" -> JsString(entity.documentation.get))
       if (entity.owner.nonEmpty) map += ("owner" -> JsString(entity.owner.get))
+      if (entity.managers.size > 0) map += ("managers" -> JsArray(entity.managers.map(JsString(_)).toVector))
```
rushtong: Use `entity.managers.nonEmpty` for consistency

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -107,7 +107,8 @@ case class AgoraEntity(namespace: Option[String] = None,
                        entityType: Option[AgoraEntityType.EntityType] = None,
                        id: Option[ObjectId] = None,
                        methodId: Option[ObjectId] = None,
-                       method: Option[AgoraEntity] = None) {
+                       method: Option[AgoraEntity] = None,
+                       managers: Seq[String] = Seq()) {
```
rushtong, line 111: Should this be an `Option[Seq[String]]`?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -36,7 +36,6 @@ abstract class AgoraService extends HttpService with RouteHelpers {
     matchNamespacePermissionsRoute(path) { (namespace, username) =>
       parameterMap { (params) =>
         val agoraEntity = AgoraEntity(Option(namespace))
-
```
rushtong, line 39: Cleaner commit without this change.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
```diff
@@ -166,7 +166,7 @@ trait QueryRouteHelper extends BaseRoute {
     val url         = params.getOrElse("url", Nil).headOption
     val _type       = params.getOrElse("entityType", Nil).headOption.toAgoraEntityOption
 
-    AgoraEntity(namespace, name, _id, synopsis, docs, owner, createDate = None, payload, url, _type)
+    AgoraEntity(namespace, name, _id, synopsis, docs, owner, createDate=None, payload, url, _type)
```
rushtong: Another non-change.

rushtong:
Looks good, works good - see minor comments in line.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
rushtong: I created https://broadinstitute.atlassian.net/browse/GAWB-1292 for this.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
skwalker: I had that at first, but @davidangb suggested just making it a seq that's set as empty instead of an option set at none

rushtong: I'm good with that - more of a stylistic thing.

rushtong:
👍

davidangb (approved):

davidangb:
:+1: thanks for sticking this out!

PR #171, opened 2016-11-17
GAWB-1236: Check user permissions before modifying ACLs

rushtong:
PR for https://broadinstitute.atlassian.net/browse/GAWB-1236

~Add a check for deletion of permissions to prevent removing *all* managers and therefore leaving the entity in an un-manageable state.~

Check to make sure the current user is not modifying their own permissions (as per updated requirements).

---
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email


MattPutnam:
👍 Seems legit to me.

src/main/resources/swagger/agora.yaml
```diff
@@ -126,7 +126,6 @@ paths:
           description: Malformed input.
           schema:
             type: object
-            code: 400
```
davidangb, line 129: thank you :)

src/test/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/ManagerPermissionSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  override def afterAll() = {
+    clearDatabases()
+  }
+
+  "Agora" should "not delete the last manager namespace permission." in {
+    agoraBusiness.insert(testEntity, mockAuthenticatedOwner.get)
+    insertNamespacePermission(testEntity, AccessControl(testEntity1.owner.get, AgoraPermissions(All)))
+    insertNamespacePermission(testEntity, AccessControl(testEntity2.owner.get, AgoraPermissions(All)))
```
davidangb: suggestion: also insert some user as a reader (non-manager), to test that additional permissions don't confuse the logic.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -191,13 +190,23 @@ trait PermissionsClient {
     }
   }
 
+  private def findRemainingManagers(agoraEntity: AgoraEntity, userToRemove: String): Set[AccessControl] = {
+    listPermissions(agoraEntity).
+      filter(_.roles.canManage).
+      filterNot(_.user.equalsIgnoreCase(userToRemove)).
+      toSet
+  }
```
davidangb: There are two places later in this class that check findRemainingManagers.size. I think it would be better encapsulated to add a new method a la `hasRemainingManagers(): Boolean` which internally checks the size. Then, calling code can just check `hasRemainingManagers` or `!hasRemainingManagers`

rushtong: 👍

rushtong:
I want to refactor `PermissionBusiness` some more - I don't like the duplicate db calls. With all of them making Await calls, I'd wager they will be a bottleneck.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
```diff
@@ -1,72 +1,114 @@
...
+    checkSameRequester(requester, userToRemove)
+    authorizeEntityRequester(getEntityACLs(entity, requester), entity, requester)
     AgoraEntityPermissionsClient.deleteEntityPermission(entity, userToRemove)
   }
 
-  def authorizeNamespaceRequester(entity: AgoraEntity, requester: String) = {
-    if (!NamespacePermissionsClient.getNamespacePermission(entity, requester).canManage &&
-        !NamespacePermissionsClient.getNamespacePermission(entity, "public").canManage)
-      throw new NamespaceAuthorizationException(AgoraPermissions(Manage), entity, requester)
+
```
rushtong, line 65: The `authorizeNamespaceRequester` and `authorizeEntityRequester` methods previously made two db queries. Each is refactored so that we can make a single db call for the relevant ACLs and pass those in.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
```diff
@@ -191,13 +190,27 @@ trait PermissionsClient {
...
+
+  private def hasRemainingManagers(agoraEntity: AgoraEntity, userToRemove: String): Boolean = {
+    findRemainingManagers(agoraEntity, userToRemove).nonEmpty
+  }
```
davidangb: we can strip all this logic out now, right? since we're checking for users attempting to edit their own ACLs.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
```diff
@@ -1,72 +1,114 @@
...
+  private def authorizeNamespaceRequester(acls: Seq[AccessControl], entity: AgoraEntity, requester: String): Unit = {
+    if (!acls.exists(_.roles.canManage))
+      throw NamespaceAuthorizationException(AgoraPermissions(Manage), entity, requester)
+  }
+
+  private def authorizeEntityRequester(acls: Seq[AccessControl], entity: AgoraEntity, requester: String): Unit = {
+    if (!acls.exists(_.roles.canManage))
+      throw AgoraEntityAuthorizationException(AgoraPermissions(Manage), entity, requester)
+  }
```
davidangb, line 103: could you remove some of the boilerplate from the business methods above (e.g. `insertEntityPermission`) and move it down into these private utility methods? For instance, if you had:
```
  private def authorizeEntityRequester(entity: AgoraEntity, requester: String): Unit = {
    authorizeEntityRequester(getEntityACLs(entity, requester), entity, requester)
  }

  private def authorizeEntityRequester(entity: AgoraEntity, requester: String, accessObject: AccessControl): Unit = {
    val currentACLs = getEntityACLs(entity, requester)
    checkSameRequesterAndPermissions(currentACLs.find(acl => acl.user.equals(requester)), accessObject)
    authorizeEntityRequester(currentACLs, entity, requester)
  }
```

then a bunch of the code up above could be simpler.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
```diff
@@ -1,72 +1,114 @@
...
+
+  private def checkSameRequester(requester: String, userToModify: String):Unit = {
+    if (requester.equalsIgnoreCase(userToModify)) {
+      val m = "Unable to remove access control for current user"
+      throw AgoraException(m, new Exception(m), Conflict)
+    }
+  }
```
davidangb, line 109: I know this is currently only called by deletes - I'd still prefer it had the generic message "Unable to modify..." to make it easier to call from elsewhere in the future.

Also, if we *really* want to make this work like rawls, rawls returns a 400 with "You may not change your own permissions". Might be worthwhile switching to that - eventually we could use the same UI components to handle them.

davidangb (commented):
comments inline!

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/PermissionsClient.scala
rushtong: I had considered that - but left it in as a safeguard. I'm fine removing it though.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
rushtong: The problem with that is we then end up making extra db calls in the other calls. The goal in this was to first get the ACLs so logic could be performed on them in two different ways. Moving the getACL calls into the authorize methods means we'll have to call them twice in some places.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
rushtong: Good call - I'll make that happen.

src/main/scala/org/broadinstitute/dsde/agora/server/business/PermissionBusiness.scala
rushtong: One more detail is that not all of the methods need `checkSameRequesterAndPermissions`, but all of them need an authorize. So, adding that into authorize will change the existing logic for the `list` and `delete` calls.

rushtong: Ignore what I wrote - I misread the suggestion. Will look at this.

davidangb (approved):
Looks good to me overall! I have one last suggestion in #173, following up my previous comment. Feel free to merge that one, or reject it, as you see fit.

I tested against a method; did not test against a config or a namespace.

rushtong:
One more bug to fix - when modifying multiple ACLs and one errors out, the others do not error out. Need to check for errors across all in the batch before applying them individually.

davidangb:
👍 the batch authorizations:
```
accessObjectList.foreach(accessObject => authorizeNamespaceRequester(entity, requester, accessObject))
```
do make multiple queries to get the entity ACLs, but I'm okay with that. You could create yet another variant of `authorizeNamespaceRequester` and `authorizeEntityRequester` to avoid the multiple queries; not sure it's worth the optimization.

PR #180, opened 2017-03-21
Slick back to 3.2 for reals this time

helgridly:


dvoet (approved):
if it actually works then ok

PR #185, opened 2017-05-18
GAWB-1784: entity snapshot permissions

davidangb:
define "entity" as "method, config, or task".

Previous behavior:
* you can create a new entity if you have Create permissions on the target namespace

After this PR, you can create a new entity:
* if non-redacted snapshots of the same namespace/name/entity type exist, and you have Manage permission on at least one of those snapshots, or
* if zero non-redacted snapshots exist, and you have Create permission on the target namespace

Note that in Agora, we consider Create and Manage to be separate permissions. However, in FireCloud, we consolidate both of those under OWNER ... so effectively, you need OWNER on a pre-existing snapshot, or on the namespace if no snapshots exist. If we ever move FireCloud away from a coarse OWNER definition, then Create/Manage will come into play.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -32,7 +32,7 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     if (snapshots.isEmpty)
       checkNamespacePermission(db, agoraEntity, username, AgoraPermissions(Create))(op)
     else {
-      val ownerPerm = AgoraPermissions(Manage)
+      val ownerPerm = AgoraPermissions(Create)
       DBIO.sequence(snapshots map { db.aePerms.getEntityPermission(_, username) }) flatMap { snapshotPermissions =>
         val nonRedacted = snapshotPermissions.filter(_.permissions > 0) // ignore redacted snapshots
         if (!nonRedacted.exists(_.hasPermission(ownerPerm)))
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -34,7 +34,8 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     else {
       val ownerPerm = AgoraPermissions(Manage)
       DBIO.sequence(snapshots map { db.aePerms.getEntityPermission(_, username) }) flatMap { snapshotPermissions =>
-        if (!snapshotPermissions.exists(_.hasPermission(ownerPerm)))
+        val nonRedacted = snapshotPermissions.filter(_.permissions > 0) // ignore redacted snapshots
+        if (!nonRedacted.exists(_.hasPermission(ownerPerm)))
           DBIO.failed(AgoraEntityAuthorizationException(ownerPerm, agoraEntity, username))
         else
           op
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.agora.server.business
 
-import org.broadinstitute.dsde.agora.server.exceptions.{AgoraEntityNotFoundException, NamespaceAuthorizationException, ValidationException}
+import org.broadinstitute.dsde.agora.server.exceptions.{AgoraEntityAuthorizationException, AgoraEntityNotFoundException, NamespaceAuthorizationException, ValidationException}
 import org.broadinstitute.dsde.agora.server.dataaccess.{AgoraDao, ReadWriteAction}
 import org.broadinstitute.dsde.agora.server.dataaccess.permissions._
 import org.broadinstitute.dsde.agora.server.dataaccess.permissions.AgoraPermissions._
@@ -24,6 +24,21 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
       }
     } else {
       DBIO.successful(AgoraPermissions(Nothing))
+    }
+  }
+  
+  private def checkInsertPermission[T](db: DataAccess, agoraEntity: AgoraEntity, username: String, snapshots: Seq[AgoraEntity])(op: => ReadWriteAction[T]): ReadWriteAction[T] = {
+    // if previous snapshots exist, check ownership on them; if not, check permission on the namespace
+    if (snapshots.isEmpty)
+      checkNamespacePermission(db, agoraEntity, username, AgoraPermissions(Create))(op)
+    else {
+      val ownerPerm = AgoraPermissions(Manage)
+      DBIO.sequence(snapshots map { db.aePerms.getEntityPermission(_, username) }) flatMap { snapshotPermissions =>
+        if (!snapshotPermissions.exists(_.hasPermission(ownerPerm)))
+          DBIO.failed(AgoraEntityAuthorizationException(ownerPerm, agoraEntity, username))
+        else
+          op
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -27,6 +27,22 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     }
   }
 
+  private def checkInsertPermission[T](db: DataAccess, agoraEntity: AgoraEntity, username: String, snapshots: Seq[AgoraEntity])(op: => ReadWriteAction[T]): ReadWriteAction[T] = {
+    // if previous snapshots exist, check ownership on them; if not, check permission on the namespace
+    if (snapshots.isEmpty)
+      checkNamespacePermission(db, agoraEntity, username, AgoraPermissions(Create))(op)
+    else {
+      val ownerPerm = AgoraPermissions(Manage)
```
aednichols: Would it be more idiomatic to use the `All` permission here? It seems like owners would be able to do anything.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -27,6 +27,22 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     }
   }
 
+  private def checkInsertPermission[T](db: DataAccess, agoraEntity: AgoraEntity, username: String, snapshots: Seq[AgoraEntity])(op: => ReadWriteAction[T]): ReadWriteAction[T] = {
```
aednichols: `op` means `operation`, right? Not imperative to change, seems to be a convention in this file

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -117,15 +133,19 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
   def insert(agoraEntity: AgoraEntity, username: String): Future[AgoraEntity] = {
-    //this goes to Mongo, so do this outside the permissions txn
+    //these next two go to Mongo, so do them outside the permissions txn
     val configReferencedMethodOpt = agoraEntity.entityType.get match {
       case AgoraEntityType.Configuration => Some(resolveMethodRef(agoraEntity.payload.get))
       case _ => None
     }
+    // find all previous snapshots. Use the dao directly, because we want to find
+    // everything, including those snapshots the current user can't read
+    val snapshots: Seq[AgoraEntity] = AgoraDao.createAgoraDao(agoraEntity.entityType)
```
aednichols: Type annotation unnecessary, can be inferred - but keep if intended for clarity

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/MockAgoraDirectives.scala
```diff
@@ -15,9 +14,15 @@ trait MockAgoraDirectives extends AgoraDirectives {
     provide(AgoraConfig.mockAuthenticatedUserEmail)
   }
 
+  // allow unit tests to specify the user making the request; if they don't, use the default
   def usernameFromRequest(magnet: ImplicitMagnet[ExecutionContext]): Directive1[String] = {
-    provide(AgoraConfig.mockAuthenticatedUserEmail)
+    optionalHeaderValueByName(MockAgoraDirectives.mockUserHeader) map {
+      case Some(mockuser) => mockuser
```
aednichols: Kinda pedantic, but this method switches between "user", "user name", and "user email"; I assume these all refer to email?

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+
+    assertResult(expectedNs2Perms) {ns2perms}
+  }
+
+  it should "allow creating a new entity if you have permissions on the namespace" in {
+    // owner1 create in namespace1
+    assertEntityCreation(owner1.get, namespace1.get, randUUID)
+  }
+
+  it should "reject creating a new entity if you lack permissions on the namespace" in {
```
aednichols, line 69: Would it be helpful to also test rejection if `owner3` tries to create an entity on `namespace1`? It's slightly different coverage, because `owner3` has never created an entity before while `owner1` has

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+    assertEntityCreation(owner1.get, namespace1.get, name)
+
+    val ent = AgoraEntity(namespace1, Some(name), Some(1))
+
+    // add owner2
+    patiently(permissionBusiness.insertEntityPermission(
+      ent, owner1.get, AccessControl(owner2.get, AgoraPermissions(All))
+    ))
+    // remove owner1
+    patiently(permissionBusiness.deleteEntityPermission(
```
aednichols: Can you explain why it's necessary to remove permission for `owner1`? It seems like the test only cares about what `owner2` can do

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+    assertEntityCreation(owner2.get, namespace1.get, name)
+  }
+
+  it should "allow creating a new snapshot of an existing entity if you own any previous snapshot of that entity" in {
+
+    val name = randUUID
+    assertEntityCreation(owner2.get, namespace2.get, name) // create snapshot 1
+    assertEntityCreation(owner2.get, namespace2.get, name) // create snapshot 2
+
+    val ent = AgoraEntity(namespace2, Some(name), Some(2))
```
aednichols: Might be nice to call this `entitySnapshot2` to more obviously indicate that `Some(2)` is specifying the second snapshot

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -117,15 +133,19 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
+      .find(AgoraEntity(agoraEntity.namespace, agoraEntity.name), None)
 
     permissionsDataSource.inTransaction { db =>
       db.aePerms.addUserIfNotInDatabase(username) flatMap { _ =>
-        checkNamespacePermission(db, agoraEntity, username, AgoraPermissions(Create)) {
+        checkInsertPermission(db, agoraEntity, username, snapshots) {
```
aednichols, line 148: Nice semantics change

aednichols (approved):
Looks good to me, just put some clarifying questions inline

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
davidangb: `op` does mean `operation` - and it is a convention throughout workbench. See, for instance:
https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
or
https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
davidangb: I specifically chose to not use `All` here. In practice, the Agora `All` permission maps to the FireCloud OWNER permission: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala#L60. Similarly, only FireCloud OWNERs will have Agora `Manage`. But if we ever went to a more granular approach, it seems like checking `Manage` is the right thing to do.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
davidangb: fixing ...

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/MockAgoraDirectives.scala
davidangb: fixing ... though since I don't want to change too much code and config files, slightly hampered ...

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
davidangb: good catch; this was not necessary. Updating the test ...

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
davidangb: Sure, good call; adding.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/EntityCreationPermissionSpec.scala
davidangb: updated!

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
davidangb: fyi, Brad asked specifically to use `Create`, not `Manage`, because it gets shown in the error message.

aednichols:
Changes LGTM, looks like CI is taking its time though

aednichols (approved):

PR #194, opened 2017-06-30
Enforce naming rules during new method creation, take 2 [GAWB-1614]

aednichols:
Testing done:

1. Swagger - creating new valid methods works, creating new invalid methods fails
2. Unit tests for valid & invalid cases
3. Local Agora + local Orchestration + local UI: methods repo loads as expected, can browse normally

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -140,6 +140,19 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     }
   }
 
+  // Name & namespace for new methods are validated against the regex.
+  // Existing methods get a pass because cleaning up the repo is prohibitively difficult (GAWB-1614)
+  private def validateNamesForNewEntity[T](entity: AgoraEntity)(op: => ReadWriteAction[T]): ReadWriteAction[T] = {
+    val nameRegex = """[a-zA-Z0-9-_.]+""".r
```
davidangb: since this is a constant, create it somewhere outside this method so we don't need to create it every time. Maybe in a companion object to AgoraBusiness.

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessTest.scala
```diff
@@ -35,6 +35,29 @@ class AgoraBusinessTest extends FlatSpec with Matchers with BeforeAndAfterAll wi
...
+      patiently(agoraBusiness.insert(testAgoraEntityWithIllegalNameChars, mockAuthenticatedOwner.get))
+    }
+  }
+
+  "Agora" should "throw an exception when inserting an entity with an illegal namespace" in {
+    intercept[ValidationException] {
+      patiently(agoraBusiness.insert(testAgoraEntityWithIllegalNamespaceChars, mockAuthenticatedOwner.get))
+    }
+  }
+
```
davidangb, line 63: I'd love to see one or more additional tests here: if an illegal name already exists in the DB, can we successfully retrieve it and/or list it? (this is the situation we ran into when the original code merged to dev)

You'd have to bypass the AgoraBusiness.insert method and use the AgoraDao directly to get the invalid name into the db, which is awkward - but it feels like an important test case.

davidangb (changes requested):
looking good, two comments inline!

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessTest.scala
aednichols: I have insert-via-business + find-via-business working, as well as insert-via-dao + find-via-dao... but I can't yet insert via dao and find via business. 

I think I'm just writing to a different mock data store than the one I'm reading from, shouldn't be too hard to figure out

aednichols: Throughout `AgoraBusiness`, we're using the `createAgoraDao` factory method to create a new DAO instance every time, so I'm guessing parameterizing the DAO and injecting its mock is not the right answer

aednichols: Idea - since Agora uses both MySQL and Mongo, there's probably a MySQL insertion required alongside Mongo to make `AgoraBusiness` `findSingle` work

davidangb (approved):
looks good, thanks for cranking through that test!

PR #195, opened 2017-07-05
GAWB-2251: endpoint to get permissions for multiple methods in one call

davidangb:
in support of 1-click sharing, create a new Agora endpoint that returns the permissions for multiple methods in one call.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
@@ -3,13 +3,16 @@ package org.broadinstitute.dsde.agora.server.webservice
 import akka.actor.ActorSystem
 import org.broadinstitute.dsde.agora.server.AgoraTestFixture
 import org.broadinstitute.dsde.agora.server.business.AgoraBusiness
-import org.broadinstitute.dsde.agora.server.AgoraTestData._
+import org.broadinstitute.dsde.agora.server.AgoraTestData.{mockAuthenticatedOwner, _}
+import org.broadinstitute.dsde.agora.server.dataaccess.permissions.EntityAccessControl
+import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
 import org.broadinstitute.dsde.agora.server.model.AgoraEntity
 import org.broadinstitute.dsde.agora.server.webservice.util.ApiUtil
 import org.scalatest.{BeforeAndAfterAll, DoNotDiscover, FlatSpec}
 import org.broadinstitute.dsde.agora.server.webservice.methods.MethodsService
 import spray.testkit.{RouteTest, ScalatestRouteTest}
 import spray.http.StatusCodes._
+import spray.httpx.SprayJsonSupport._
 
 import scala.concurrent.duration._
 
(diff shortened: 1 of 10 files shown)


src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
```diff
@@ -246,4 +249,48 @@ class PermissionIntegrationSpec extends FlatSpec with RouteTest with ScalatestRo
...
+        }
+        // check second - it exists, but we don't have permissions to see it
+        {
+          val stubEntity = AgoraEntity(agoraEntity2.namespace, agoraEntity2.name, agoraEntity2.snapshotId)
+          val found = entityAclList.find(_.entity == stubEntity)
+          assert(found.isDefined)
+          assert(found.get.message.nonEmpty)
+          assert(found.get.acls.isEmpty)
+        }
+        // check third - it doesn't exist in the db
```
aednichols, line 289: Would it be worthwhile sense to test a redacted entity? The behavior should be the same as for a non-existent entity, I would think.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
```diff
@@ -246,4 +249,48 @@ class PermissionIntegrationSpec extends FlatSpec with RouteTest with ScalatestRo
...
+
+    Post(ApiUtil.Methods.withLeadingVersion + "/permissions", payload) ~>
+      methodsService.multiEntityPermissionsRoute ~>
+      check {
+        assert(status == OK)
+        val entityAclList = responseAs[Seq[EntityAccessControl]]
+        assertResult(3) {entityAclList.size}
+
+        // check first - should get permissions
+        {
```
aednichols: I like the standalone scope braces, I think this is the first time I've seen someone doing this in practice

aednichols (commented):
Looks good, the response format is intuitive.

Some questions/issues besides inline comments:
- If you have an unshared method and share it with one reader, the reader's role appears as expected. There also appears an ACL entry for `public` with no roles. Why?
- Omitting one of the required fields in `configurations`/`methods` triggers error that's not very helpful:
```
{
  "code": 500,
  "message": "None.get"
}
```

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
davidangb: I've added a redacted entity to the test. Under the covers, redaction removes all permissions from the entity, so it's actually the same as the second test - it exists in the db, but we're not allowed to see it.

davidangb:
I've added better validation for when the user fails to supply a name/namespace/snapshotId.

The "public" permission is a weird quirk. I've seen that before, too. It's certainly not a result of this PR (which is read-only to the db) ... something we might want to fix at some point, even though it's harmless.

Back to you @aednichols for re-review!

aednichols (approved):
👍

PR #201, opened 2017-07-28
Changing validations so errors appear in Orchestration

UrsaStutsman:


src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -126,7 +126,6 @@ case class AgoraEntity(namespace: Option[String] = None,
 
   AgoraEntity.validate(this) match {
     case Success(_) => this
-    case Failure(errors) => throw new IllegalArgumentException(s"Entity is not valid: Errors: $errors")
```
dvoet: why not remove the call to validate entirely? Seems like just removing the error check makes this a no op anyway

UrsaStutsman: Good point. Will do.

davidangb:
I'm a bit worried about this change. By removing the call to `validate` from the `AgoraEntity` constructor, we've removed validation from a bunch of places. For instance, the POST /api/v1/methods/{namespace}/{name}/{snapshotId} endpoint has lost validation.

This also has multiple unit test failures - no idea why Travis says they're passing!! **Update:** see #203 for unit test/Travis issues

davidangb:
( assigned myself so I can watch this more easily )

davidangb:
Unit tests are still failing.

Both POST endpoints for methods are now validating, thank you!

It feels awkward to have the validation business logic encoded in three places. For instance, `AddRouteHelper.validatePostRoute`, `QuerySingleHelper.validateNewEntity`, and `AgoraEntity.validate` all code for the synopsis limit of 80 characters. Could we define the validation logic in one place, and then reference that logic from all the places that need it?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -106,6 +106,13 @@ object AgoraEntity {
     (namespace |@| name |@| _id |@| synopsis |@| doc) {(namespace, name, _id, synopsis, doc) => doNothing }
   }
 
+  def getErrors(entity: AgoraEntity): Option[IllegalArgumentException] = {
```
MatthewBemis: it feels weird that only tests are using this because then we're testing code that isn't actually ever used in production

PR #206, opened 2017-08-30
Selectively return payload as an object instead of string [GAWB-2468]

aednichols:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,8 @@ libraryDependencies ++= Seq(
   "mysql" % "mysql-connector-java" % "5.1.35",
   "org.broadinstitute" %% "wdl4s" % wdl4sV,
   "org.broadinstitute.dsde.vault" %% "vault-common" % "0.1-15-bf74315",
+  "org.broadinstitute.dsde" %% "rawls-model" % "0.1-8c63c48-SNAP"
+    exclude("com.typesafe.scala-logging", "scala-logging_2.11"),
   "org.mongodb" %% "casbah" % "2.8.2",
   "org.scalaz" %% "scalaz-core" % "7.1.3",
   "org.webjars" % "swagger-ui"  % "2.2.5",
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -163,7 +163,8 @@ paths:
           in: query
           description: Boolean to return only the payload of the method.
           required: false
-          type: string
+          type: boolean
+          default: false
       tags:
         - Methods
       responses:
@@ -439,6 +440,12 @@ paths:
           description: Snapshot ID of method.
           required: true
           type: number
+        - name: payloadAsObject
+          in: query
+          description: Instead of returning a string under key payload, return a JSON object under key payloadObject
+          required: false
+          type: boolean
+          default: false
       tags:
         - Configurations
       responses:
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
@@ -10,7 +10,6 @@ import org.joda.time.format.{DateTimeFormatter, ISODateTimeFormat}
 import spray.json.{JsArray, JsString, _}
 
 import org.broadinstitute.dsde.rawls.model.MethodConfiguration
-import org.broadinstitute.dsde.rawls.model.AttributeString._
 import org.broadinstitute.dsde.rawls.model.AttributeString
 import org.broadinstitute.dsde.rawls.model.MethodRepoMethod
 
@@ -104,6 +103,7 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
       val owner = stringOrNone(jsObject, "owner")
       val createDate = if (jsObject.getFields("createDate").nonEmpty) jsObject.fields("createDate").convertTo[Option[DateTime]] else None
       val payload = stringOrNone(jsObject, "payload")
+      val payloadObject = if (jsObject.getFields("payloadObject").nonEmpty) jsObject.fields("payloadObject").convertTo[Option[MethodConfiguration]] else None
       val url = stringOrNone(jsObject, "url")
       val entityType = if (jsObject.getFields("entityType").nonEmpty) jsObject.fields("entityType").convertTo[Option[AgoraEntityType.EntityType]] else None
       val id = if (jsObject.getFields("_id").nonEmpty) jsObject.fields("_id").convertTo[Option[ObjectId]] else None
@@ -118,6 +118,7 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
                                owner = owner,
                                createDate = createDate,
                                payload = payload,
+                               payloadObject = payloadObject,
                                url = url,
                                entityType = entityType,
                                id = id,
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
@@ -10,8 +10,7 @@ import org.joda.time.format.{DateTimeFormatter, ISODateTimeFormat}
 import spray.json.{JsArray, JsString, _}
 
 import org.broadinstitute.dsde.rawls.model.MethodConfiguration
-import org.broadinstitute.dsde.rawls.model.AttributeString
-import org.broadinstitute.dsde.rawls.model.MethodRepoMethod
+import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.MethodConfigurationFormat
 
 import scala.language.implicitConversions
 
@@ -174,19 +173,7 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
 
   implicit val AccessControlPairFormat = jsonFormat3(EntityAccessControl)
 
-  implicit val MethodRepoMethodFormat = jsonFormat3(MethodRepoMethod)
-
-  // AttributeString is just a case class with one String in it, but for some reason jsonFormat1 is not sufficient
-  // https://stackoverflow.com/a/28530361/818054
-  implicit val AttributeStringFormat = jsonFormat1(AttributeString)
-  implicit object attributeStringJsonFormat extends RootJsonFormat[AttributeString] {
-    override def read(value: JsValue) = AttributeString(value.convertTo[String])
-    override def write(f: AttributeString) = JsString(f.value)
-  }
-
-  implicit val MethodConfigurationFormat = jsonFormat10(MethodConfiguration)
-
-    implicit object AgoraStatusFormat extends RootJsonFormat[AgoraStatus] {
+  implicit object AgoraStatusFormat extends RootJsonFormat[AgoraStatus] {
     override def write(obj: AgoraStatus): JsObject = {
       JsObject("status" -> JsString(if (obj.up) "up" else "down"),
         "message" -> JsArray(obj.messages.map(JsString(_)).toVector)
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
@@ -3,11 +3,16 @@ package org.broadinstitute.dsde.agora.server.model
 
 import org.broadinstitute.dsde.agora.server.exceptions.AgoraException
 import org.broadinstitute.dsde.agora.server.webservice.util.AgoraOpenAMClient.UserInfoResponse
-import org.broadinstitute.dsde.agora.server.dataaccess.permissions.{AccessControl, EntityAccessControl, AgoraPermissions}
+import org.broadinstitute.dsde.agora.server.dataaccess.permissions.{AccessControl, AgoraPermissions, EntityAccessControl}
 import org.bson.types.ObjectId
 import org.joda.time.DateTime
 import org.joda.time.format.{DateTimeFormatter, ISODateTimeFormat}
 import spray.json.{JsArray, JsString, _}
+
+import org.broadinstitute.dsde.rawls.model.MethodConfiguration
+import org.broadinstitute.dsde.rawls.model.AttributeString._
+import org.broadinstitute.dsde.rawls.model.AttributeString
+import org.broadinstitute.dsde.rawls.model.MethodRepoMethod
 
 import scala.language.implicitConversions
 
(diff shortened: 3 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -108,27 +108,29 @@ abstract class AgoraService(permissionsDataSource: PermissionsDataSource) extend
...
-          } else {
-            entity(as[AgoraEntity]) { newEntity =>
-              AgoraEntity.validate(newEntity) match {
-                case Success(_) => //noop
-                case Failure(errors) => throw new IllegalArgumentException(s"Method is invalid: Errors: $errors")
-              }
-              parameters("redact".as[Boolean] ? false) { redact =>
-                requestContext => completeEntityCopy(requestContext, targetEntity, newEntity, redact, username, path, queryHandlerProps)
+      extractOnlyPayloadParameter { (onlyPayload: Option[String]) =>
+        extractPayloadAsObjectParameter { (payloadAsObject: Option[String]) =>
```
rushtong: In the yaml, it's defined as a boolean. Is it intentional to unmarshal it as a string here?

src/main/resources/swagger/agora.yaml
```diff
@@ -439,6 +440,12 @@ paths:
           description: Snapshot ID of method.
           required: true
           type: number
+        - name: payloadAsObject
+          in: query
+          description: Instead of returning a string under key payload, return a JSON object under key payloadObject
```
rushtong, line 445: Would be nice to have some more info saying that you can't have `payloadAsObject` and `onlyPayload` both set to true in the same request.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
```diff
@@ -58,10 +60,22 @@ class QueryHandler(dataSource: PermissionsDataSource, implicit val ec: Execution
...
+        RequestComplete(BadRequest, "onlyPayload, payloadAsObject cannot be used together")
+      } else {
+
+        if (onlyPayload) {
+          RequestComplete(foundEntity.payload)
+        } else if (payloadAsObject) {
+          RequestComplete(foundEntity.deserializeConfigurationPayload)
+        } else {
+          RequestComplete(foundEntity)
+        }
```
rushtong: These nested ifs seem like a good case for a sequence of case statements

rushtong: Perhaps something like this reads cleaner:
```
      (onlyPayload, payloadAsObject) match {
        case (o, p) if o && p => RequestComplete(BadRequest, "onlyPayload, payloadAsObject cannot be used together")
        case (o, p) if o => RequestComplete(foundEntity.payload)
        case (o, p) if p => RequestComplete(foundEntity.deserializeConfigurationPayload)
        case _ => RequestComplete(foundEntity)
      }
```

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
rushtong: The extract methods are pretty rigid and don't let you do any type coercion. I think you can replace the nested `extract...` methods with something like this:
```
      parameters( 'onlyPayload ? false, 'payloadAsObject ? false) { (onlyPayload, payloadAsObject) =>
        val targetEntity = AgoraEntity(Option(namespace), Option(name), Option(snapshotId))
        get { requestContext =>
          completeWithPerRequest(requestContext, targetEntity, username, onlyPayload, payloadAsObject, path, queryHandlerProps)
        } ~
```
See http://spray.io/documentation/1.2.4/spray-routing/parameter-directives/parameters/

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -160,5 +162,46 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
...
+  "Agora" should "not let you use payloadAsObject and onlyPayload at the same time" in {
+    Get(ApiUtil.Configurations.withLeadingVersion + "/" +
+      testConfig1.namespace.get + "/" +
+      testConfig1.name.get + "/" +
+      testConfig1.snapshotId.get + "?payloadAsObject=true&onlyPayload=true") ~>
+      configurationsService.querySingleRoute ~> check {
+      assert(body.asString contains "onlyPayload, payloadAsObject cannot be used together")
+      assert(status == BadRequest)
+    }
+  }
```
rushtong: I think it would be good to have a test with non-standard values that should coerce to false, such as `payloadAsObject=Fred&onlyPayload=Flintstone`

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -160,5 +162,46 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
     }
   }
 
+  "Agora" should "return the payload as an object if you ask it to" in {
```
rushtong: If you nest these tests, you could set a base url that can be reused by all of the tests with different query params concatenated:
```
ApiUtil.Configurations.withLeadingVersion + "/" +
      testConfig1.namespace.get + "/" +
      testConfig1.name.get + "/" +
      testConfig1.snapshotId.get
```
for DRY

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -467,6 +496,28 @@ object AgoraTestData {
     entityType = Option(AgoraEntityType.Workflow)
   )
 
+  val testMethod1 = new AgoraEntity(
```
rushtong: Nit: These are only slightly different from the other test entities in that they have a snapshot. Can you rename them to something more reflective of that? Maybe `testSnapshotMethod1` and `testSnapshotConfig1`?

src/main/resources/swagger/agora.yaml
davidangb: how about return a 400 if both are specified? (Update: ah, I see you're already doing that)

rushtong (changes requested):
Some minor feedback inline.
Separately, calls to current configs are failing for me with the payload option set to true:
```
{
  "code": 500,
  "message": "Object is missing required member 'methodConfigVersion'"
}
```
I suspect this is due to the new Rawls `MethodConfiguration` class having a required `methodConfigVersion` but existing/older configs don't have it. I think it might need to be an `Option[Int]` instead and default to 1.

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -167,7 +174,19 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
...
+  // AttributeString is just a case class with one String in it, but for some reason jsonFormat1 is not sufficient
+  // https://stackoverflow.com/a/28530361/818054
+  implicit val AttributeStringFormat = jsonFormat1(AttributeString)
+  implicit object attributeStringJsonFormat extends RootJsonFormat[AttributeString] {
+    override def read(value: JsValue) = AttributeString(value.convertTo[String])
+    override def write(f: AttributeString) = JsString(f.value)
+  }
+
+  implicit val MethodConfigurationFormat = jsonFormat10(MethodConfiguration)
+
```
davidangb: the rawls model library already has json formats defined for `MethodRepoMethod`, `MethodConfiguration`, and `AttributeString`.

Look for:
`WorkspaceModel.MethodStoreMethodFormat`
`WorkspaceModel.MethodConfigurationFormat`
`JsonSupport.AttributeStringFormat`

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -157,6 +165,23 @@ case class AgoraEntity(namespace: Option[String] = None,
...
+
+  def deserializeConfigurationPayload: AgoraEntity = {
+    if (!canDeserializePayload) throw AgoraException(s"Entity type $entityType does not support payload deserialization")
+
+    payload match {
+      case Some(pl: String) =>
+        this.copy(
+          payloadObject = Some(pl.parseJson.convertTo[MethodConfiguration]),
+          payload = None
+        )
```
davidangb: I think this wants some error-trapping/handling around the parseJson.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
davidangb: looks like these two could be simplified into a single spray directive, something like:
`parameters('onlyPayload, 'payloadAsObject) { (onlyPayload.as[Boolean], payloadAsObject.as[Boolean]) => `

and then remove the calls to `toBool` later. This would lose the error-handling/default in `toBool` but I'm okay with making the route stricter - if someone specifies something other than a boolean in these params, I think they *should* get an error instead of a default false/

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
```diff
@@ -58,10 +60,22 @@ class QueryHandler(dataSource: PermissionsDataSource, implicit val ec: Execution
...
-      if (onlyPayload) RequestComplete(foundEntity.payload)
-      else RequestComplete(foundEntity)
+      if (onlyPayload && payloadAsObject) {
+        RequestComplete(BadRequest, "onlyPayload, payloadAsObject cannot be used together")
+      } else {
+
+        if (onlyPayload) {
+          RequestComplete(foundEntity.payload)
+        } else if (payloadAsObject) {
+          RequestComplete(foundEntity.deserializeConfigurationPayload)
```
davidangb: rename `deserializeConfigurationPayload` to `withDeserializedPayload` or similar? Seeing it here parallel to `foundEntity.payload` is confusing.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -160,5 +162,46 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
...
+      testConfig1.snapshotId.get) ~>
+      configurationsService.querySingleRoute ~>
+    check {
+      assert(status == OK)
+
+      val entity = responseAs[AgoraEntity]
+      assert(entity.payloadObject.isEmpty)
+      assert(entity.payload.get contains testConfig1.payload.get)
+    }
+  }
```
davidangb: worthwhile adding a test for what happens if the user specifies "?payloadAsObject=somethingThatIsNotABoolean"

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -160,5 +162,46 @@ class AgoraConfigurationsSpec extends ApiServiceSpec {
...
+    Get(ApiUtil.Configurations.withLeadingVersion + "/" +
+      testConfig1.namespace.get + "/" +
+      testConfig1.name.get + "/" +
+      testConfig1.snapshotId.get + "?payloadAsObject=true") ~>
+      configurationsService.querySingleRoute ~>
+    check {
+      assert(status == OK)
+
+      val entity = responseAs[AgoraEntity]
+      assert(entity.payloadObject.get.isInstanceOf[MethodConfiguration])
```
davidangb: worth comparing the payloadObject to what you expect it to be

davidangb (commented):
comments inline!

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
aednichols: How does
```
(onlyPayload, payloadAsObject) match {
  case (true, true) => RequestComplete(BadRequest, "onlyPayload, payloadAsObject cannot be used together")
  case (true, false) => RequestComplete(foundEntity.payload)
  case (false, true) => RequestComplete(foundEntity.deserializeConfigurationPayload)
  case _ => RequestComplete(foundEntity)
}
```
strike you? Lets the case statement handle the boolean logic for us.

src/main/resources/swagger/agora.yaml
aednichols: It's not possible for a user to do that in Swagger because the flags apply to different endpoints:

`payloadAsObject` is only available on
```
/api/v1/configurations/{namespace}/{name}/{snapshotId}
```
and `onlyPayload` is only available on
```
/api/v1/methods/{namespace}/{name}/{snapshotId}
```
A person writing their own script will not be limited by the Swagger interface, and so would hit the 400 instead.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
aednichols: The Swagger type just defines what kind of control gets rendered for that parameter. For optional booleans, it's a dropdown limited to the empty string, `true`, and `false`.

The DSL syntax you suggested is awesome, gladly adopting!

src/main/resources/swagger/agora.yaml
rushtong: Ah - thank you!

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -167,7 +174,19 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
 
   implicit val AccessControlPairFormat = jsonFormat3(EntityAccessControl)
 
-  implicit object AgoraStatusFormat extends RootJsonFormat[AgoraStatus] {
+  implicit val MethodRepoMethodFormat = jsonFormat3(MethodRepoMethod)
+
+  // AttributeString is just a case class with one String in it, but for some reason jsonFormat1 is not sufficient
```
jmthibault79: `jsonFormat1` would translate `AttributeString("this.hello")` as `{ "value": "this.hello" }` but we want simply `"this.hello"`

build.sbt
```diff
@@ -37,6 +37,8 @@ libraryDependencies ++= Seq(
   "mysql" % "mysql-connector-java" % "5.1.35",
   "org.broadinstitute" %% "wdl4s" % wdl4sV,
   "org.broadinstitute.dsde.vault" %% "vault-common" % "0.1-15-bf74315",
+  "org.broadinstitute.dsde" %% "rawls-model" % "0.1-8c63c48-SNAP"
+    exclude("com.typesafe.scala-logging", "scala-logging_2.11"),
```
davidangb, line 41: could we use a non-snapshot version instead? I just checked orch and realized it uses a snapshot version - and that's probably where you got this from - which sucks. Orch shouldn't be doing that either.

davidangb (commented):
getting close; would still like to see:
* some error-trapping around the payload deserialization, to handle pre-existing malformed payloads
* some tests for non-boolean values in `payloadAsObject`, `onlyPayload` parameters

davidangb (approved):
👍 thanks for the updates ... looks good, works good!

build.sbt
aednichols: How do I specify the most recent, non-snapshot version?

PR #209, opened 2017-09-07
Handle old configs with missing methodConfigVerison [GAWB-2468]

aednichols:
This is a workaround for a spray-json limitation - it is not able to read & use default values specified by case classes.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
@@ -79,27 +79,20 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
     // Mirror the default values in the MethodConfiguration case class - spray-json does not know how to use them
     // https://stackoverflow.com/questions/15740925/what-is-a-good-way-to-handle-default-values-with-spray-json
     override def read(json: JsValue): MethodConfiguration = {
-      // get the fields once so we don't do extra work
-      val fields = json.asJsObject.fields
-
-      // check required keys
-      val requiredKeys = Set("namespace","name","rootEntityType","prerequisites","inputs","outputs","methodRepoMethod")
-      val missingKeys = requiredKeys diff fields.keySet
-      if (missingKeys.nonEmpty)
-        throw DeserializationException(s"Failed to read field(s) [${missingKeys.mkString(",")}] from method configuration")
+      val jsObject = json.asJsObject
 
       MethodConfiguration(
-        namespace = fields("namespace").convertTo[String],
-        name = fields("name").convertTo[String],
-        rootEntityType = fields("rootEntityType").convertTo[String],
-        prerequisites = fields("prerequisites").convertTo[Map[String, AttributeString]],
-        inputs = fields("inputs").convertTo[Map[String, AttributeString]],
-        outputs = fields("outputs").convertTo[Map[String, AttributeString]],
-        methodRepoMethod = fields("methodRepoMethod").convertTo[MethodRepoMethod],
+        namespace = getFieldOrFail[String](jsObject, "namespace"),
+        name = getFieldOrFail[String](jsObject, "name"),
+        rootEntityType = getFieldOrFail[String](jsObject, "rootEntityType"),
+        prerequisites = getFieldOrFail[Map[String, AttributeString]](jsObject, "prerequisites"),
+        inputs = getFieldOrFail[Map[String, AttributeString]](jsObject, "inputs"),
+        outputs = getFieldOrFail[Map[String, AttributeString]](jsObject, "outputs"),
+        methodRepoMethod = getFieldOrFail[MethodRepoMethod](jsObject, "methodRepoMethod"),
 
-        methodConfigVersion = fields.getOrElse("methodConfigVersion",JsNumber(1)).convertTo[Int],
-        deleted = fields.getOrElse("deleted",JsBoolean(false)).convertTo[Boolean],
-        deletedDate = fields.get("deletedDate") map (_.convertTo[DateTime])
+        methodConfigVersion = getFieldOrDefault(jsObject, "methodConfigVersion", 1),
+        deleted = getFieldOrDefault(jsObject, "deleted", false),
+        deletedDate = getFieldOrDefault[Option[DateTime]](jsObject, "deletedDate", None)
       )
     }
   }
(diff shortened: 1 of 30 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -69,6 +71,50 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
...
+      map += ("rootEntityType" -> JsString(obj.rootEntityType))
+      map += ("prerequisites" -> obj.prerequisites.toJson)
+      map += ("inputs" -> obj.inputs.toJson)
+      map += ("outputs" -> obj.outputs.toJson)
+      map += ("methodRepoMethod" -> obj.methodRepoMethod.toJson)
+      map += ("methodConfigVersion" -> JsNumber(obj.methodConfigVersion))
+      map += ("deleted" -> JsBoolean(obj.deleted))
+      if (obj.deletedDate.nonEmpty) map += ("deletedDate" -> JsString(obj.deletedDate.get.toString()))
+
+      JsObject(map)
```
davidangb: looks like there is no special logic in the `write` method here. Would
```
    override def write(obj: MethodConfiguration): JsValue =
      jsonFormat10(MethodConfiguration).write(obj)
```
work just as well?

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -69,6 +71,50 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
...
+          throw DeserializationException(s"Failed to read field 'inputs' from method configuration"),
+        outputs = if (jsObject.getFields("outputs").nonEmpty) jsObject.fields("outputs").convertTo[Map[String, AttributeString]] else
+          throw DeserializationException(s"Failed to read field 'outputs' from method configuration"),
+        methodRepoMethod = if (jsObject.getFields("methodRepoMethod").nonEmpty) jsObject.fields("methodRepoMethod").convertTo[MethodRepoMethod] else
+          throw DeserializationException(s"Failed to read field 'methodRepoMethod' from method configuration"),
+        methodConfigVersion = if (jsObject.getFields("methodConfigVersion").nonEmpty) jsObject.fields("methodConfigVersion").convertTo[Int] else 1,
+        deleted = if (jsObject.getFields("deleted").nonEmpty) jsObject.fields("deleted").convertTo[Boolean] else false,
+        deletedDate = if (jsObject.getFields("deletedDate").nonEmpty) jsObject.fields("deletedDate").convertTo[Option[DateTime]] else None
+      )
+    }
```
davidangb: I also rewrote the `read` method to be a little less verbose and be a bit more Scala-y:

```
    // Mirror the default values in the MethodConfiguration case class - spray-json does not know how to use them
    // https://stackoverflow.com/questions/15740925/what-is-a-good-way-to-handle-default-values-with-spray-json
    override def read(json: JsValue): MethodConfiguration = {
      val jsObject = json.asJsObject

      // get the fields once so we don't do extra work
      val fields = jsObject.fields

      // check required keys
      val requiredKeys = Set("namespace","name","rootEntityType","prerequisites","inputs","outputs","methodRepoMethod")
      val missingKeys = requiredKeys diff fields.keySet
      if (missingKeys.nonEmpty)
        throw DeserializationException(s"Failed to read field(s) [${missingKeys.mkString(",")}] from method configuration")

      MethodConfiguration(
        namespace = fields("namespace").convertTo[String],
        name = fields("name").convertTo[String],
        rootEntityType = fields("rootEntityType").convertTo[String],
        prerequisites = fields("prerequisites").convertTo[Map[String, AttributeString]],
        inputs = fields("inputs").convertTo[Map[String, AttributeString]],
        outputs = fields("outputs").convertTo[Map[String, AttributeString]],
        methodRepoMethod = fields("methodRepoMethod").convertTo[MethodRepoMethod],

        methodConfigVersion = fields.getOrElse("methodConfigVersion",JsNumber(1)).convertTo[Int],
        deleted = fields.getOrElse("deleted",JsBoolean(false)).convertTo[Boolean],
        deletedDate = fields.get("deletedDate") map (_.convertTo[DateTime])
      )
    }
```

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraConfigurationsSpec.scala
```diff
@@ -222,6 +222,22 @@ class AgoraConfigurationsSpec extends ApiServiceSpec with FlatSpecLike {
...
+      Get(baseURL + "?payloadAsObject=true") ~>
+      configurationsService.querySingleRoute ~> check {
+        assert(status == OK)
+
+        val entity = responseAs[AgoraEntity]
+
+        assert(entity.payloadObject.get.methodConfigVersion == 1)
+      }
+    }
+
```
davidangb, line 241: you could even have a test here along the lines of:
```
    "Agora" should "throw DeserializationError if missing keys" in {
      import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport.MethodConfigurationFormat
      val ex = intercept[DeserializationException] {
        MethodConfigurationFormat.read(JsObject())
      }
      assertResult("Failed to read field(s) [name,methodRepoMethod,outputs,inputs,rootEntityType,prerequisites,namespace] from method configuration") {
        ex.getMessage
      }
    }
```

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
aednichols: Yesterday I made a couple functions that would do a lot to cut repetitiveness, but it wasn't immediately obvious how to handle spray's demand for a JSON formatter for type `T`
```
private def getFieldOrDefault[T](json: JsObject, key: String, default: T): T = {
  if (json.getFields(key).nonEmpty) json.fields(key).convertTo[T] else default
}
private def getFieldOrFail[T](json: JsObject, key: String): T = {
  if (json.getFields(key).nonEmpty) json.fields(key).convertTo[T] else 
    throw DeserializationException(s"Deserialization failed due to missing field $key")
}
```

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraApiJsonSupport.scala
```diff
@@ -69,6 +71,32 @@ object AgoraApiJsonSupport extends DefaultJsonProtocol {
...
+        prerequisites = getFieldOrFail[Map[String, AttributeString]](jsObject, "prerequisites"),
+        inputs = getFieldOrFail[Map[String, AttributeString]](jsObject, "inputs"),
+        outputs = getFieldOrFail[Map[String, AttributeString]](jsObject, "outputs"),
+        methodRepoMethod = getFieldOrFail[MethodRepoMethod](jsObject, "methodRepoMethod"),
+
+        methodConfigVersion = getFieldOrDefault(jsObject, "methodConfigVersion", 1),
+        deleted = getFieldOrDefault(jsObject, "deleted", false),
+        deletedDate = getFieldOrDefault[Option[DateTime]](jsObject, "deletedDate", None)
+      )
+    }
```
davidangb: honestly, I like this version less than the version in #332bf0. This is for a few reasons:

1. The previous version gave you an error message with all the missing keys; the current version will short-circuit with an exception on the first missing key. Once you fix the first missing key, you'll suddenly get an exception on the second missing key.

2. The previous boilerplate was actually shorter:
```
old: fields("name").convertTo[String]
new: getFieldOrFail[String](jsObject, "name")

old: fields.get("deletedDate") map (_.convertTo[DateTime])
new: getFieldOrDefault[Option[DateTime]](jsObject, "deletedDate", None)
```

3. I think getting the jsObject `fields` once and using it is better than calling `getFields()` repeatedly, because `getFields` is just slightly more work under the covers. This is a micro-optimization and should not at all be the deciding factor.

davidangb (approved):
👍

PR #211, opened 2017-09-11
swagger tweaks

davidangb:
view this PR with `?w=1` in the url to ignore whitespace changes; it's much cleaner.

adds/removes `type: object` in the right places to get `ConfigurationPayload` and `methodRepoMethod` to show up in the swagger ui.


aednichols (approved):
This is great. We do
```
type: object
$ref: '#/foo/bar'
```
a lot and I thought that might be as good as it got.

PR #212, opened 2017-09-19
GAWB-2601: unauthenticated ga4gh endpoint for public WDL

davidangb:
This PR adds an unauthenticated, GA4GH-compliant (mostly) endpoint to get the WDL for a public method snapshot. This is a building block for allowing WDL imports during Cromwell execution.

This will need an accompanying orchestration passthrough.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -1315,11 +1315,6 @@ paths:
           schema:
             type: object
             $ref: '#/definitions/Error'
-      security:
-        - authorization:
-            - openid
-            - email
-            - profile
   /ga4gh/tools/{id}/versions/{version-id}/{type}/descriptor:
     get:
       summary: Get the tool descriptor (CWL/WDL) for the specified tool.
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -1406,6 +1406,52 @@ paths:
             - openid
             - email
             - profile
+  /ga4gh/tools/{id}/versions/{version-id}/{type}/descriptor:
+    get:
+      summary: Get the tool descriptor (CWL/WDL) for the specified tool.
+      description: Returns the CWL or WDL descriptor for the specified tool.
+      tags:
+        - GA4GH
+      parameters:
+        - name: type
+          required: true
+          in: path
+          description: >
+            The output type of the descriptor. If not specified it is up to the underlying
+            implementation to determine which output type to return. Plain types return the
+            bare descriptor while the "non-plain" types return a descriptor wrapped
+            with metadata.
+            In Agora, only WDL and plain-WDL are supported.
+          type: string
+          enum:
+            - CWL
+            - WDL
+            - plain-CWL
+            - plain-WDL
+        - name: id
+          in: path
+          description: >
+            A unique identifier of the tool, scoped to this registry, for example `123456`.
+            In Agora, this must be a namespace + ":" + name. For instance, if your namespace
+            is 'foo' and name is 'bar', this must be 'foo:bar'.
+          required: true
+          type: string
+        - name: version-id
+          in: path
+          required: true
+          type: string
+          description: >
+            An identifier of the tool version for this particular tool registry, for example `v1`.
+            In Agora, this must be an integer representing the Agora snapshot id.
+      responses:
+        '200':
+          description: The tool descriptor.
+          schema:
+            $ref: '#/definitions/ToolDescriptor'
+        '404':
+          description: The tool can not be output in the specified type.
+          schema:
+            $ref: '#/definitions/Error'
 securityDefinitions:
   authorization:
     type: oauth2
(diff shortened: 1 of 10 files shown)


src/main/resources/swagger/agora.yaml
```diff
@@ -1406,6 +1406,52 @@ paths:
...
+            In Agora, this must be an integer representing the Agora snapshot id.
+      responses:
+        '200':
+          description: The tool descriptor.
+          schema:
+            $ref: '#/definitions/ToolDescriptor'
+        '404':
+          description: The tool can not be output in the specified type.
+          schema:
+            $ref: '#/definitions/Error'
```
davidangb, line 1366: I kept the swagger definitions as close as possible to spec: https://github.com/ga4gh/tool-registry-schemas/blob/develop/src/main/resources/swagger/ga4gh-tool-discovery.yaml. I just added a few "In Agora ... " lines of description.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
```diff
@@ -0,0 +1,38 @@
...
+          get { requestContext =>
+
+            val dType = parseDescriptorType(descriptorType)
+            val snapshotId = parseVersionId(versionId)
+            val toolId = parseId(id)
+
+            val entity = AgoraEntity(Some(toolId.namespace), Some(toolId.name), Some(snapshotId), entityType = Some(AgoraEntityType.Workflow))
+
+            val message = QueryPublicSingle(requestContext, entity, dType)
+            perRequest(requestContext, queryHandler, message)
```
davidangb, line 32: I think that calling `perRequest` right from the route - and using the Spray DSL to define the route - is easier to understand than the current agora archetype. So I did it this way, instead of following the pattern I don't like.

I also kept all (most) ga4gh-related code in a single directory, am totally willing to hear negative opinions about that choice.

src/main/resources/swagger/agora.yaml
```diff
@@ -1406,6 +1406,52 @@ paths:
...
+            In Agora, only WDL and plain-WDL are supported.
+          type: string
+          enum:
+            - CWL
+            - WDL
+            - plain-CWL
+            - plain-WDL
+        - name: id
+          in: path
+          description: >
```
helgridly, line 1343: curiosity Q: how is `>` different to `|`?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
```diff
@@ -84,6 +90,22 @@ class QueryHandler(dataSource: PermissionsDataSource, implicit val ec: Execution
     }
   }
 
+  def queryPublic(requestContext: RequestContext, entity: AgoraEntity, descriptorType: ToolDescriptorType.DescriptorType): Future[PerRequestMessage] = {
+    val entityTypes = Seq(entity.entityType.getOrElse(throw new ValidationException("need an entity type")))
+    agoraBusiness.findSingle(entity, entityTypes, AccessControl.publicUser) map { foundEntity =>
+
+      descriptorType match {
+        case ToolDescriptorType.WDL =>
+          val result = ToolDescriptor(foundEntity.url.getOrElse(""),
```
davidangb: The urls returned by agora are known to be incorrect; this is GAWB-1741. I have chosen to continue to return incorrect urls here, like we do elsewhere. Once GAWB-1741 is fixed, these urls will magically be correct.

The other option is to always return the empty string for url, and once GAWB-1741 is fixed, we'd need to come back here to add the url back in. Reviewer: which is better?

src/main/resources/swagger/agora.yaml
davidangb: Good answer [here](https://stackoverflow.com/questions/15540635/what-is-the-use-of-pipe-symbol-in-yaml/18708156#18708156) ... `|` preserves newlines, `>` ignores them ... except for indented blocks.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
aednichols: I think we should return the URLs as-is but note it as a known issue somewhere (comment or Swagger).

The incorrect URLs are easily made useful if you fix the domain, as compared to the empty string.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
aednichols: I endorse both choices.

Less boilerplate is good.

And there's so little GA4GH stuff that it would get lost if it was broken up by function rather than by domain.

aednichols (approved):

PR #218, opened 2017-10-11
GAWB-849: ga4gh api for agora.

davidangb:
I'm pushing this PR out now for initial review, because it's already gotten large and unwieldy. I'd like to get this preliminary PR merged, then do follow-on work to clean it up. Because this is almost all additive and hidden, I think this is safe.

Must do before merging: a corresponding orchestration PR to reflect that the single ga4gh endpoint exposed by orch has moved, from /ga4gh/ to /ga4gh/v1/. I'll update here once I have that PR.

What this PR does:
* adds ga4gh tool-registry spec (https://github.com/ga4gh/tool-registry-schemas) endpoints for everything that agora can support
* populates the ga4gh model with the minimum necessary data to claim a working endpoint
* copy-and-paste of swagger from the ga4gh spec into agora
* unit tests

What is left to do:
* update swagger to reflect what we _actually_ support, not what the spec supports
* orch passthroughs for all endpoints
* populate the models more fully
* additional unit tests, as appropriate

This google doc: https://docs.google.com/a/broadinstitute.org/spreadsheets/d/1QkgDMlVobQC9t5PjNqMg1nv80IusOm_ovfW7HHmCeGU/edit?usp=sharing
has a summary of progress and level of compatibility for endpoints and models.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Inform other teams of any API changes via Slack and/or email

### DIFF
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -1649,6 +1649,20 @@ securityDefinitions:
       openid: open id authorization
       email: email authorization
       profile: profile authorization
+
+parameters:
+  limit:
+      name: limit
+      in: query
+      description: Amount of records to return in a given page.  By default it is 1000.
+      type: integer
+      format: int32
+  offset:
+      name: offset
+      in: query
+      description: Start index of paging. Pagination results can be based on numbers or other values chosen by the registry implementor (for example, SHA values). If this exceeds the current result set return an empty set.  If not specified in the request this will start at the beginning of the results.
+      type: string
+
 definitions:
   EntityId:
     type: object
@@ -2138,6 +2152,17 @@ definitions:
       url:
         type: string
         description: 'Optional url to the tool descriptor used to build this image, should include version information, and can include a git hash (e.g. https://raw.githubusercontent.com/ICGC-TCGA-PanCancer/pcawg_delly_workflow/ea2a5db69bd20a42976838790bc29294df3af02b/delly_docker/Delly.cwl )'
+  ToolTests:
+    description: A tool document that describes how to test with one or more sample test JSON.
+    required:
+      - test
+    properties:
+      test:
+        type: string
+        description: The test JSON content for this tool.
+      url:
+        type: string
+        description: 'Optional url to the test JSON used to test this tool'
   ToolDockerfile:
     description: A tool dockerfile is a document that describes how to build a particular Docker image.
     required:
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -1322,7 +1322,133 @@ paths:
           schema:
             type: object
             $ref: '#/definitions/Error'
-  /ga4gh/tools/{id}/versions/{version-id}/{type}/descriptor:
+
+  /ga4gh/v1/tools/{id}:
+    get:
+      summary: List one specific tool, acts as an anchor for self references
+      description: This endpoint returns one specific tool (which has ToolVersions nested inside it)
+      tags:
+        - GA4GH
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+          description: A unique identifier of the tool, scoped to this registry, for example `123456`
+      responses:
+        '200':
+          description: A tool.
+          schema:
+              $ref: '#/definitions/Tool'
+      security: []
+
+  /ga4gh/v1/tools/{id}/versions:
+    get:
+      summary: List versions of a tool
+      description: Returns all versions of the specified tool
+      tags:
+        - GA4GH
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+          description: A unique identifier of the tool, scoped to this registry, for example `123456`
+      responses:
+        '200':
+          description: An array of tool versions
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/ToolVersion'
+      security: []
+
+  /ga4gh/v1/tools/{id}/versions/{version-id}:
+    get:
+      summary: List one specific tool version, acts as an anchor for self references
+      description: This endpoint returns one specific tool version
+      tags:
+        - GA4GH
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+          description: A unique identifier of the tool, scoped to this registry, for example `123456`
+        - name: version-id
+          in: path
+          required: true
+          type: string
+          description: An identifier of the tool version, scoped to this registry, for example `v1`
+      responses:
+        '200':
+          description: A tool version.
+          schema:
+              $ref: '#/definitions/ToolVersion'
+      security: []
+
+  /ga4gh/v1/tools:
+    get:
+      summary: List all tools
+      description: >
+        This endpoint returns all tools available or a filtered subset using
+        metadata query parameters.
+      tags:
+        - GA4GH
+      parameters:
+        - name: id
+          type: string
+          in: query
+          description: A unique identifier of the tool, scoped to this registry, for example `123456`
+        - name: registry
+          in: query
+          type: string
+          description: The image registry that contains the image.
+        - name: organization
+          in: query
+          type: string
+          description: The organization in the registry that published the image.
+        - name: name
+          in: query
+          type: string
+          description: The name of the image.
+        - name: toolname
+          in: query
+          type: string
+          description: The name of the tool.
+        - name: description
+          in: query
+          type: string
+          description: The description of the tool.
+        - name: author
+          in: query
+          type: string
+          description: 'The author of the tool (TODO a thought occurs, are we assuming that the author of the CWL and the image are the same?).'
+        - $ref: '#/parameters/offset'
+        - $ref: '#/parameters/limit'
+      responses:
+        '200':
+          description: An array of Tools that match the filter.
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Tool'
+          headers:
+            next-page:
+              description: A URL that can be used to reach the next page based on the current offset and page record limit
+              type: string
+            last-page:
+              description: A URL that can be used to reach the last page based on the current page record limit
+              type: string
+            current-offset:
+              description: The current start index of the paging used for this result
+              type: string
+            current-limit:
+              description: The current page record limit used for this result
+              type: integer
+      security: []
+
+  /ga4gh/v1/tools/{id}/versions/{version-id}/{type}/descriptor:
     get:
       summary: Get the tool descriptor (CWL/WDL) for the specified tool.
       description: Returns the CWL or WDL descriptor for the specified tool.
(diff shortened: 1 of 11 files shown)


src/main/resources/swagger/agora.yaml
```diff
@@ -1322,7 +1322,133 @@ paths:
           schema:
             type: object
             $ref: '#/definitions/Error'
-  /ga4gh/tools/{id}/versions/{version-id}/{type}/descriptor:
+
```
davidangb, line 1325: please read PR description - I know some of this swagger is incorrect and needs updating.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
```diff
@@ -3,36 +3,79 @@ package org.broadinstitute.dsde.agora.server.ga4gh
...
+        path("tools" / Segment / "versions" / Segment / Segment / "descriptor") { (id, versionId, descriptorType) => requestContext =>
+          val entity = entityFromArguments(id, versionId)
+          val message = QueryPublicSinglePayload(requestContext, entity, parseDescriptorType(descriptorType))
+          perRequest(requestContext, queryHandler, message)
+        } ~
+        path("tools" / Segment / "versions" / Segment / Segment / "descriptor" / Segment) { (id, versionId, descriptorType, relativePath) =>
+          complete(spray.http.StatusCodes.NotImplemented)
+        } ~
+        path("tools" / Segment / "versions" / Segment / Segment / "tests") { (id, versionId, descriptorType) =>
+          complete(spray.http.StatusCodes.NotImplemented)
```
davidangb, line 77: I realize these routes could have been written more hierarchically, e.g. 
```
pathPrefix("tools") {
  pathEnd {} ~
  pathPrefix(Segment) {
    pathEnd {} ~
    pathPrefix("versions") {
      ...
    }
  }
}
```
but after going down that route, I rewrote it "flat" as you can see above; I thought it was more readable.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
```diff
@@ -13,6 +112,20 @@ object Models {
     val PLAIN_CWL = Value("plain-CWL")
   }
 
+  // ToolTests
+
+  case class ToolDockerfile(
+    dockerfile: String,
+    url: String)
```
davidangb, line 119: will delete this model if I can't get its endpoint working in the future

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
```diff
@@ -1,10 +1,109 @@
...
+        description="", // TODO: should we use a snapshot synopsis or description?
+        author="", // TODO: can we get this from the WDL?
+        `meta-version`="", // TODO
+        contains=List.empty[String],
+        verified=false,
+        `verified-source`="",
+        signed=false,
+        versions=entities.toList map (x => ToolVersion(x))
+      )
+    }
```
davidangb, line 70: there's a decent bit of business logic in these `apply` methods. It might make more sense to have e.g.:
```
object Tool {
  def apply(entities:Seq[AgoraEntity]) = {
    toolFromEntities(entities)
  }
}
```
and then break out the `toolFromEntities` into some other class. Thoughts?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/handlers/QueryHandler.scala
```diff
@@ -98,25 +92,6 @@ class QueryHandler(dataSource: PermissionsDataSource, implicit val ec: Execution
...
-          val result = ToolDescriptor(foundEntity.url.getOrElse(""),
-            foundEntity.payload.getOrElse(""),
-            ToolDescriptorType.WDL)
-          RequestComplete(result)
-        case ToolDescriptorType.PLAIN_WDL =>
-          RequestComplete(foundEntity.payload.getOrElse(""))
-      }
-    }
-  }
-
```
davidangb, line 119: I moved all the ga4gh-specific handlers into their own dedicated class.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/util/ServiceMessages.scala
```diff
@@ -76,10 +76,6 @@ object ServiceMessages {
                          onlyPayload: Boolean,
                          payloadAsObject: Boolean)
 
-  case class QueryPublicSingle(requestContext: RequestContext,
-                         entity: AgoraEntity,
-                         descriptorType: DescriptorType)
-
```
davidangb, line 82: I moved all the ga4gh-specific messages into their own dedicated class.

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestSuite.scala
```diff
@@ -28,7 +28,9 @@ class AgoraTestSuite extends Suites(
   new PermissionIntegrationSpec,
   new MethodDefinitionIntegrationSpec,
   new AssociatedConfigurationIntegrationSpec,
+  new CompatibleConfigurationIntegrationSpec,
```
davidangb, line 31: this has nothing to do with the current PR, but, uh, we were not running those tests at all.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
rushtong: 👍  for legibility.

src/main/resources/swagger/agora.yaml
rushtong: This didn't validate for me - I needed to add a `parameters` and `ToolTests` section (which directly mimics the GA4GH spec). I have a PR to this PR with that change: https://github.com/broadinstitute/agora/pull/219

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
rushtong: Need to know if we should `TODO` the unimplemented endpoints, or even if they are implementable with what we have.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
rushtong: 👍  I like the idea of some kind of `ModelSupport` class that does the heavy lifting and leaving the case classes in their own little world.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
```diff
@@ -3,36 +3,79 @@ package org.broadinstitute.dsde.agora.server.ga4gh
...
-            val toolId = parseId(id)
-
-            val entity = AgoraEntity(Some(toolId.namespace), Some(toolId.name), Some(snapshotId), entityType = Some(AgoraEntityType.Workflow))
-
-            val message = QueryPublicSingle(requestContext, entity, dType)
-            perRequest(requestContext, queryHandler, message)
+    pathPrefix("ga4gh" / "v1") {
+      get {
+        path("metadata") {
+          // because of the dashes in these names, using the automatic spray json marshalling is annoying.
```
rushtong, line 26: Something to consider is using a different json marshaller, if possible, something like circe which I think the green team is using. Definitely scope creep, but if we run into more difficulties with spray-json, we should consider it.

src/test/scala/org/broadinstitute/dsde/agora/server/model/Ga4ghModelTest.scala
```diff
@@ -0,0 +1,154 @@
...
+          `descriptor-type` = List("WDL"),
+          dockerfile = false,
+          `meta-version` = "3",
+          verified = false,
+          `verified-source` = "")
+        val actual = ToolVersion(defaultEntity)
+      }
+    }
+    // TODO: 
+    "ToolDescriptor" - {}
```
rushtong, line 149: `    "ToolTests" - {}`

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
```diff
@@ -13,6 +112,20 @@ object Models {
     val PLAIN_CWL = Value("plain-CWL")
   }
 
+  // ToolTests
```
rushtong, line 115: Need a TODO for this.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
```diff
@@ -1,10 +1,109 @@
...
+    def apply(entity:AgoraEntity) = fromEntityType(entity.entityType)
+    def apply(method:MethodDefinition) = fromEntityType(method.entityType)
+    def fromEntityType(entityType: Option[AgoraEntityType.EntityType]) = {
+      assert(entityType.nonEmpty, "cannot create a ToolClass if entity type is empty")
+      val str = entityType.get.toString
+      new ToolClass(str, str, "")
+    }
+  }
+
+  case class Tool(
```
rushtong, line 37: Some of these fields are not defined in the GA4GH model, but I'm guessing that's OK? My guess is that as long as we support the minimum, anything extra is gravy. This is also where we'll need alignment with the swagger spec.

rushtong (commented):
👍  Looks good. The endpoints worked as expected. Several minor comments inline, but mostly just clarifications. I have one PR that I'll submit against this for the swagger yaml so it validates.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
davidangb: agreed. I also figured out how to do this with the spray-json, just by using backticks in the model:
```
  case class Metadata(
    version: String,
    `api-version`: String,
    country: String,
    `friendly-name`: String)
```
but I never updated the comment or the endpoint.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Models.scala
davidangb: which ones are not defined in the GA4GH model? I tried to be exact: https://github.com/ga4gh/tool-registry-schemas/blob/develop/src/main/resources/swagger/ga4gh-tool-discovery.yaml#L335

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghService.scala
davidangb: I don't think we are able to implement the unimplemented endpoints - Agora just doesn't support those things. The relative-path descriptor and tests endpoints are definite no-gos. We could *possibly* implement the dockerfile endpoint, but that would require parsing the WDL and pulling out the info ... but even then, within a Workflow, each Task might have a different dockerfile, and so which should we report?

So, my plan was:
* remove swagger defs for these endpoints, so we only show swagger for what we actually support
* remove models, tests, TODOs, and any code that references these endpoints
* BUT, leave the route in place and respond with NotImplemented (instead of NotFound) just in case a power user attempts the route - this is a signal that we know the route should exist per the spec, but we don't have anything there.

PR #221, opened 2017-10-18
GAWB-849: Refine supported GA4GH APIs

rushtong:
Partially addresses https://broadinstitute.atlassian.net/browse/GAWB-849

## Changes
* Remove unimplemented endpoints (and unused models) from Swagger
* Flesh out model classes as much as possible.
* More tests

## Notes for reviewer:
* Goes hand in hand with https://github.com/broadinstitute/firecloud-orchestration/pull/517, but can be reviewed separately. 
* I made an executive decision on the definition of `organization`  as `"The Broad Institute of Harvard and MIT"` and we likely need PO approval for this.
* Question: should I add some documentation around empty fields that agora doesn't support?
* Question: do we need swagger documentation around empty fields that agora doesn't support?
  - Specifically, these fields are not supported in Agora:
  - Tool:contains
  - Tool:verified
  - Tool:verified-source
  - Tool:signed
  - ToolVersion:dockerfile
  - ToolVersion:image
  - ToolVersion:verified
  - ToolVersion:verified-source

---
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/Ga4ghQueryHandler.scala
```diff
@@ -17,6 +17,9 @@ import scala.concurrent.{ExecutionContext, Future}
 class Ga4ghQueryHandler(dataSource: PermissionsDataSource, override implicit val ec: ExecutionContext)
   extends QueryHandler(dataSource, ec) {
 
+  // Include entity payloads in returned results
+  val DefaultProjection = Option(new AgoraEntityProjection(Seq[String]("payload", "synopsis"), Seq.empty[String]))
```
davidangb: `Some(AgoraEntityProjection ...` instead

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+  * Support class to handle apply methods with some degree of business logic
+  */
+object ModelSupport {
+
+  final val organization = "The Broad Institute of Harvard and MIT"
+  final val verifiedSource = ""
+  final val version = "1.0.0"
+  final val apiVersion = version
```
davidangb: while I don't anticipate changing these anytime soon, these versions should not be tied together. We could update our version before the spec is updated.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def toolClassFromEntityType(entityType: Option[AgoraEntityType.EntityType]): ToolClass = {
+    assert(entityType.nonEmpty, "cannot create a ToolClass if entity type is empty")
+    val str = entityType.get.toString
+    new ToolClass(str, str, "")
+  }
+
+  def toolFromEntities(entities:Seq[AgoraEntity]): Tool = {
+    val representative = entities.last
+    val versions = entities.toList map (x => ModelSupport.toolVersionFromEntity(x))
```
davidangb: can remove the `ModelSupport.` prefix

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+    new Tool(
+      url=url,
+      id=id,
+      organization=organization,
+      toolname=latestVersion.name,
+      toolclass=ToolClass(representative),
+      description=representative.synopsis.getOrElse(""),
+      author=author,
+      `meta-version` = latestVersion.`meta-version`,
+      contains=List.empty[String],
```
davidangb: Since you're already parsing the WDL ... I had considered populating the `contains` field with the names of the tasks contained in the WDL. What do you think?

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+      versions=versions
+    )
+  }
+
+  def toolVersionFromEntity(entity: AgoraEntity): ToolVersion = {
+    assert(entity.entityType.contains(AgoraEntityType.Workflow), "cannot create a ToolVersion if entityType is not Workflow")
+    assert(entity.snapshotId.nonEmpty, "cannot create a ToolVersion if snapshot id is empty")
+    new ToolVersion(
+      name = entity.name.getOrElse(""),
+      url = entity.url.getOrElse(""),
```
davidangb: you could use the same approach as `AgoraConfig.GA4GH.toolUrl` here.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+      verified = false,
+      `verified-source` = verifiedSource
+    )
+  }
+
+  def toolDescriptorFromEntity(entity: AgoraEntity): ToolDescriptor = {
+    new ToolDescriptor(
+      `type` = ToolDescriptorType.WDL,
+      descriptor = entity.payload.getOrElse(""),
+      url = entity.url.getOrElse("")
```
davidangb: you could use the same approach as `AgoraConfig.GA4GH.toolUrl` here.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+    )
+  }
+
+  /**
+    * Metadata is always a constant.
+    */
+  def metadata(): Metadata = {
+    new Metadata(version = version, `api-version` = apiVersion, country = country, `friendly-name` = friendlyName)
```
davidangb: don't need `new`

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,120 @@
...
+    val emailField = "email"
+    val stringFormat = "%s <%s>"
+    payload match {
+      case Some(wdl) =>
+        WdlNamespaceWithWorkflow.load(wdl, Seq.empty) match {
+          case Success(parsed) =>
+            val authorEmails: Seq[(String, String)] = {
+              parsed.tasks.map(_.meta.getOrElse(authorField, "")).zip(parsed.tasks.map(_.meta.getOrElse(emailField, ""))) ++
+                parsed.workflows.map(_.meta.getOrElse(authorField, "")).zip(parsed.workflows.map(_.meta.getOrElse(emailField, "")))
+            }.filterNot(_._1.isEmpty).distinct
```
davidangb: I love this approach of pulling stuff out of the WDL.

If I read this correctly, the logic is:
* if WDL developer specified author and email, use "author \<email\>"
* if WDL developer specified only author, use "author"
* if WDL developer specified only email, ignore.

We could handle the last case with code something like:
```
parsed.tasks.map { task =>
                val email = task.meta.getOrElse(emailField,"").trim
                val author = task.meta.getOrElse(authorField,"").trim
                val coalescedAuthor = if (author.nonEmpty) author else email
                author -> email
              }
```

davidangb (commented):
Some comments inline, nothing huge. Remaining work is updating the individual swagger details to show our implementation - for instance, if we always return "" for a field we should say so.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
rushtong: You read that correctly, and I like that additional case ... wish I had thought of it :-)

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
rushtong: I like that ... will dig into the spec and see what we can get.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
rushtong: The whole thing could be much simpler by handling the string interpolation inside the map as well.

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  final val organization = "The Broad Institute of Harvard and MIT"
+  final val verifiedSource = ""
+  final val version = "1.0.0"
+  final val apiVersion = "1.0.0"
+  final val country = "USA"
+  final val friendlyName = "FireCloud"
+
+  def toolIdFromEntity(entity:AgoraEntity): ToolId = {
+    assert(entity.namespace.nonEmpty, "cannot create a ToolId if entity namespace is empty")
```
aednichols, line 23: These asserts are a good idea, I wonder why name & namespace aren't encoded in the type system with `String` instead of `Option[String]`

aednichols (commented):
Looks good, endpoints work - except some weirdness with one of them.

The `/ga4gh/v1/tools` endpoint is returning `The server was not able to produce a timely response to your request.` when I query with all-blank filter fields.

Querying that same endpoint with `id` value  `mbemis:mb-test` yields error 
```
{
  "code": 500,
  "message": "ERROR: Unexpected symbol (line 163, col 25) when parsing '_gen15'.\n\nExpected rbrace, got ,.\n\n\t\tMutectTask.MAFLiteFile,\n                        ^\n\n$wf_output_wildcard_syntax = :fqn $_gen16 -> WorkflowOutputWildcard( fqn=$0, wildcard=$1 )\n     ",
  "timestamp": 1508465294121
}
```
It could be that there's a problem with Matt's method, but the blank query should still work.

aednichols:
Querying `/ga4gh/v1/tools` with `id` value `mbemis:mb-test` no longer returns an error, but it does not seem to actually apply the filter - all tools are returned.

rushtong:
@aednichols - I'll look at that. Did you get extra results through the agora swagger or through the orch swagger? I'll check both though. I didn't change the agora handling, so this is puzzling.

aednichols:
@rushtong this was with orch swagger

src/main/resources/swagger/agora.yaml
```diff
@@ -1395,63 +1395,19 @@ paths:
         metadata query parameters.
       tags:
         - GA4GH
-      parameters:
```
davidangb: remove "or a filtered subset" from the description of this endpoint

src/main/resources/swagger/agora.yaml
```diff
@@ -2077,27 +1904,13 @@ definitions:
...
-           type: string
-      verified:
-        type: boolean
-        description: Reports whether this tool has been verified by a specific organization or individual
-      verified-source:
-        type: string
-        description: Source of metadata that can support a verified tool, such as an email or URL
-      signed:
-        type: boolean
-        description: Reports whether this tool has been signed.
```
davidangb: I don't think you should remove these properties. We do return these keys in the json responses; they're just constantly blank/false. Either leave them as-is defined by the spec, or feel free to add e.g. "always returns empty string" for our implementation.

Update: same issue for other models below. Bah, maybe this means you need to update orch swagger too; I didn't look closely enough over there!

davidangb (commented):
looking good, a few comments inline

src/main/resources/swagger/agora.yaml
rushtong: They are basically the same - I'm trying to make the same changes in both places.

davidangb (approved):
:+1:

aednichols (approved):

PR #223, opened 2017-10-25
GAWB-849: allow for both plain text and application/json in the swagger UI.

rushtong:
The GA4GH descriptor endpoint can produce either plain text or json, so we need to allow for that in the swagger UI

---

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/agora.yaml
+++ b/src/main/resources/swagger/agora.yaml
@@ -1439,6 +1439,9 @@ paths:
             An identifier of the tool version for this particular tool registry, for example `v1`.
 
               *In Agora, this must be an integer representing the Agora snapshot id.*
+      produces:
+        - application/json
+        - text/plain
       responses:
         '200':
           description: The tool descriptor.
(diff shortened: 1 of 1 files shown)


aednichols (approved):
On dev changing from `WDL` to `plain-WDL` "just works", but on this branch that's an error unless you also change the content type. This seems like it could annoy users, so we should probably add a note about it somewhere

davidangb (commented):
:+1: thanks for fixing!

the ga4gh spec defines this `produces:` at a global level: https://github.com/ga4gh/tool-registry-schemas/blob/develop/src/main/resources/swagger/ga4gh-tool-discovery.yaml#L6, which we shouldn't do - so I think this is a good fix. Do we need this for orch too?

PR #240, opened 2018-02-26
GAWB-3052: Admin google group polling (skip tests)

kyuksel:
Uncommenting out admin google group poller and moving it to `ServerInitializer` since `ApiService` is no longer an `Actor` and polling the group is not an Agora api service.


rushtong (approved):
👍 LGTM

PR #251, opened 2018-03-07
Gr gawb 3052 cleanup

rushtong:
Miscellaneous cleanup tasks:
* Exception handling was somewhat dropped and temporarily different from what was done in PerRequest. Updated it to work like it did there. 
* Go through remaining TODOs in the code and address or remove. 
* Fixed a minor logging bug in `AgoraException`

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
@@ -64,9 +64,7 @@ abstract class AgoraService(permissionsDataSource: PermissionsDataSource) extend
               val deletionAttempt = agoraBusiness.delete(targetEntity, entityTypes, username)
 
               onComplete(deletionAttempt) {
-                case Success(rowsDeleted) => complete("FIX ME")
-                // TODO[GAWB-3052] The line above should be the line below; figure out what implicit may be missing
-                // case Success(rowsDeleted) => complete(rowsDeleted)
+                case Success(rowsDeleted) => complete(rowsDeleted.toString)
                 case Failure(error) => failWith(error)
               }
             } ~
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiService.scala
@@ -35,10 +35,10 @@ object ApiService extends LazyLogging with SprayJsonSupport with DefaultJsonProt
   val exceptionHandler: ExceptionHandler = {
     ExceptionHandler {
       case withErrorReport: WorkbenchExceptionWithErrorReport =>
-        complete(withErrorReport.errorReport.statusCode.getOrElse(StatusCodes.InternalServerError), withErrorReport.errorReport)
+        complete(withErrorReport.errorReport.statusCode.getOrElse(InternalServerError), withErrorReport.errorReport)
       case workbenchException: WorkbenchException =>
-        val report = ErrorReport(Option(workbenchException.getMessage).getOrElse(""), Some(StatusCodes.InternalServerError), Seq(), Seq(), Some(workbenchException.getClass))
-        complete(StatusCodes.InternalServerError, report)
+        val report = ErrorReport(Option(workbenchException.getMessage).getOrElse(""), Some(InternalServerError), Seq(), Seq(), Some(workbenchException.getClass))
+        complete(InternalServerError, report)
       case e: IllegalArgumentException => complete(BadRequest, e)
       case e: AgoraEntityAuthorizationException => complete(Forbidden, AgoraException(e.getMessage, e.getCause, Forbidden))
       case e: NamespaceAuthorizationException => complete(Forbidden, AgoraException(e.getMessage, e.getCause, Forbidden))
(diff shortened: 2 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
```diff
@@ -64,9 +64,7 @@ abstract class AgoraService(permissionsDataSource: PermissionsDataSource) extend
               val deletionAttempt = agoraBusiness.delete(targetEntity, entityTypes, username)
 
               onComplete(deletionAttempt) {
-                case Success(rowsDeleted) => complete("FIX ME")
-                // TODO[GAWB-3052] The line above should be the line below; figure out what implicit may be missing
-                // case Success(rowsDeleted) => complete(rowsDeleted)
+                case Success(rowsDeleted) => complete(rowsDeleted.toString)
```
rushtong, line 67: There is no marshaller for `Int`, but the original code was calling `toString` on the rows value.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiService.scala
```diff
@@ -37,17 +39,19 @@ object ApiService {
...
+      case e: AgoraEntityNotFoundException => complete(NotFound, AgoraException(e.getMessage, e.getCause, NotFound))
+      case e: DockerImageNotFoundException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: PermissionNotFoundException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: ValidationException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: PermissionModificationException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: AgoraException => complete(StatusCodes.getForKey(e.statusCode.intValue).getOrElse(StatusCodes.InternalServerError), e.getMessage)
+      case e: wdl.exception.ValidationException => complete(BadRequest, e.getMessage)
+      case e: Throwable =>
+        logger.error("Exception caught by ExceptionHandler: ", e)
+        complete(StatusCodes.InternalServerError, e.getMessage)
```
rushtong: Compare these to the original `PerRequest` error handling.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/RouteHelpers.scala
```diff
@@ -18,8 +18,6 @@ trait BaseRoute extends RouteUtil  {
       throw new IllegalArgumentException("Missing params: user and/or role.")
   }
 
-  // TODO[GAWB-3052] L used to be a subtype of HList. Akka-Http dropped shapeless.
-  // Does L have to be a subtype of something else (e.g. Product) now?
```
rushtong, line 22: No need to do anything here.

aednichols (approved):

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiService.scala
```diff
@@ -37,17 +39,19 @@ object ApiService {
...
+      case e: AgoraEntityNotFoundException => complete(NotFound, AgoraException(e.getMessage, e.getCause, NotFound))
+      case e: DockerImageNotFoundException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: PermissionNotFoundException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: ValidationException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: PermissionModificationException => complete(BadRequest, AgoraException(e.getMessage, e.getCause, BadRequest))
+      case e: AgoraException => complete(e.statusCode, e.getMessage)
+      case e: wdl.exception.ValidationException => complete(BadRequest, e.getMessage)
+      case e: Throwable =>
+        logger.error("Exception caught by ExceptionHandler: ", e)
+        complete(StatusCodes.InternalServerError, e.getMessage)
```
davidangb: these last three cases that complete with a string message - do they return json? If not, can we wrap those messages in an `AgoraException` and return json?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
davidangb: I'm pretty sure there's no `RootJsonFormat` for `Int`, even though there's a `JsonFormat` for it (and the *Formats get turned into *Marshallers eventually). That's because a raw int isn't a valid json packet - it needs to be inside an object or an array.  So, for instance, `complete(Seq(rowsDeleted))` would work - but that's not what we want to return.

i.e. the business logic is wrong here, not the marshallers. I'm fine with using `toString` as a workaround.

davidangb (approved):
:+1: one potential improvement inline

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraService.scala
rushtong: If the business logic is wrong here, then it was wrong in the original code as well. I suppose we could story this topic for an improvement at some point if we're all OK with this approach.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/ApiService.scala
rushtong: The last three are actually not corrected as they should be ... they should be completing with an exception, not a string message. Will fix.

davidangb (approved):
:+1:

PR #253, opened 2018-03-08
PR feedback

davidangb:
PR feedback for #252 - see comments inline!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ organization := "org.broadinstitute"
 scalaVersion := "2.12.4"
 
 // uncomment to see compile warnings
-// scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
+scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
 
 lazy val akkaV = "2.5.9"
 lazy val akkaHttpV = "10.0.11"
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/SwaggerRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/SwaggerRoutes.scala
@@ -6,8 +6,6 @@ import akka.http.scaladsl.server.Directives._
 import akka.stream.scaladsl.Flow
 import akka.util.ByteString
 
-import scala.language.postfixOps
-
 trait SwaggerRoutes {
   private val swaggerUiPath = "META-INF/resources/webjars/swagger-ui/2.2.10-1"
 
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -8,7 +8,7 @@ organization := "org.broadinstitute"
 scalaVersion := "2.12.4"
 
 // uncomment to see compile warnings
-// scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
+scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
```
davidangb, line 11: now that you've fixed all the deprecation warnings, let's enable this!

src/main/scala/org/broadinstitute/dsde/agora/server/SwaggerRoutes.scala
```diff
@@ -31,8 +29,8 @@ trait SwaggerRoutes {
       // want to catch all paths lest we circumvent Spray's not-found and method-not-allowed error
       // messages.
       (pathSuffixTest("o2c.html") | pathSuffixTest("swagger-ui.js")
-        | pathPrefixTest("css" /) | pathPrefixTest("fonts" /) | pathPrefixTest("images" /)
-        | pathPrefixTest("lang" /) | pathPrefixTest("lib" /)) {
+        | pathPrefixTest("css") | pathPrefixTest("fonts") | pathPrefixTest("images")
+        | pathPrefixTest("lang") | pathPrefixTest("lib")) {
```
davidangb, line 33: the warning about `postfixOps` was due to these trailing slashes ... which aren't necessary afaict. Instead of fixing the warning by importing `postfixOps`, I fixed the warning by removing the slashes.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/mongo/EmbeddedMongo.scala
```diff
@@ -11,7 +11,7 @@ object EmbeddedMongo extends MongoEmbedDatabase {
   def startMongo() = {
     println(s"Starting embedded mongo db instance.")
     if (mongodProps == null || !mongodProps.mongodProcess.isProcessRunning) {
-      mongodProps = mongoStart(port = 27017, version = Version.V3_4_1)
+      mongodProps = mongoStart(port = 27017, version = Version.V3_0_6)
```
davidangb, line 14: in prod, our mongo version is 3.0.6, so let's match that here. I realize this gives a deprecation warning, but in this particular case I think that's good ... it'll prod us to upgrade (GAWB-3037)

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestSuite.scala
```diff
@@ -32,6 +32,7 @@ class AgoraTestSuite extends Suites(
   new CompatibleConfigurationIntegrationSpec,
   new Ga4ghServiceSpec,
   new Ga4ghModelTest,
+  new SwaggerRoutesSpec,
```
davidangb, line 35: while investigating the trailing slashes in swagger routes, I wrote some tests (see below)

rushtong (approved):
👍 Love it!

PR #238, opened 2018-02-20
GAWB-3052: Migration to Akka-Http

kyuksel:
### Goal
Address [GAWB-3052](https://broadinstitute.atlassian.net/browse/GAWB-3052) by replacing `Spray` with `Akka-Http` so `Agora` can be upgraded to `Scala 2.12`

### Status
- [x] App booting (`Agora` => `ServerInitializer` => `ApiService`)
- [x] `StatusService`
- [x] `SwaggerRoutes`
- [x] Admin Google Group synchronization ([PR](https://github.com/broadinstitute/agora/pull/240))
- [x] `Ga4ghService` ([PR](https://github.com/broadinstitute/agora/pull/241))
- [x] Remove the remaining three `Spray` imports from `ApiServiceSpec2` ([PR](https://github.com/broadinstitute/agora/pull/244))
- [x] Graceful exit ([PR](https://github.com/broadinstitute/agora/pull/242))
- [x] Replace `ApiServiceActor.supervisorStrategy` ([PR](https://github.com/broadinstitute/agora/pull/243))
- [x] `AgoraService.postRoute` ([PR](https://github.com/broadinstitute/agora/pull/244))
- [x] `AgoraService.queryRoute` ([PR](https://github.com/broadinstitute/agora/pull/245))
- [x] `AgoraService.queryCompatibleConfigurationsRoute` ([PR](https://github.com/broadinstitute/agora/pull/247))
- [x] `AgoraService.queryAssociatedConfigurationsRoute` ([PR](https://github.com/broadinstitute/agora/pull/247))
- [x] `AgoraService.queryMethodDefinitionsRoute` ([PR](https://github.com/broadinstitute/agora/pull/247))
- [x] `AgoraService.multiEntityPermissionsRoute` ([PR](https://github.com/broadinstitute/agora/pull/246))
- [x] `AgoraService.entityPermissionsRoute` ([PR](https://github.com/broadinstitute/agora/pull/246))
- [x] `AgoraService.namespacePermissionsRoute` ([PR](https://github.com/broadinstitute/agora/pull/246))
- [x] Make sure all tests are turned back on in `AgoraTestSuite`
- [x] Remove the depracated files
- [x] Remove the remaining (non-`spray.json`) `spray` references from: ([PR](https://github.com/broadinstitute/agora/pull/248))
   - `DockerHubClient`
   - `DockerHubJsonSupport`
   - `AgoraBusiness`
   - `AgoraException`
[...]

PR #255, opened 2018-03-15
GAWB-3052: Request completion refactoring

kyuksel:
- Addresses the PR feedback for #238 
- Removes redundant `Future` `Failure` handling in request `complete`ions where we passed `Throwable`s through without modification. 
- Removes `onComplete`s where above applies and the content of `Future` `Success` is not further processed.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -463,8 +463,45 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     findSingle(entity.namespace.get, entity.name.get, entity.snapshotId.get, entityTypes, username)
   }
 
+//  private def validateDockerImage(task: Task) = {
+//    // Per DSDEEPB-2525, in the interests of expediency, we are disabling docker image validation as we do not support validation of private docker images
+//    // When that functionality is added back in, then this is where it should go.
+//
+////    if (task.runtimeAttributes.docker.isDefined) {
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -463,45 +463,8 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
-//
-////    if (task.runtimeAttributes.docker.isDefined) {
-////      val dockerImageReference = parseDockerString(task.runtimeAttributes.docker.get)
-////      if (dockerImageReference.isDefined) {
-////        DockerHubClient.doesDockerImageExist(dockerImageReference.get)
-////      }
-////    }
-//    true
-//  }
-
```
rushtong: Minor nit - I understand the deletion here, but I would prefer that happen not as part of the migration to 2.12 and instead as something else. Same for the deleted block below this.

rushtong (approved):
👍 Mostly good. Would it make sense to add all of the base PR feedback to this PR? It would be nice to do this in one go.

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
kyuksel: Understood. They're back in now.

davidangb (commented):
akka-http has a `completeWith` directive, so could you choose a different name to avoid confusion? Otherwise 👍 !

davidangb (approved):
:+1:

rushtong (approved):
👍

PR #257, opened 2018-03-27
GAWB-3057 Upgrade Agora dependencies

kyuksel:
- Addresses [GAWB-3057](https://broadinstitute.atlassian.net/browse/GAWB-3057)
- [Recommended diff](https://github.com/broadinstitute/agora/compare/ky_GAWB-3052_spray_migration...ky_GAWB-3057_assess_dependencies?diff=split&name=ky_GAWB-3057_assess_dependencies) from the yet-unmerged [Scala 2.12 PR](https://github.com/broadinstitute/agora/pull/238)
- `swagger-ui` and `mockserver-netty` were not upgraded as the newer versions break and it didn't feel worth the time to make them work.


build.sbt
```diff
@@ -19,36 +22,36 @@ resolvers += "artifactory-releases" at artifactory + "libs-release"
 resolvers += "artifactory-snapshots" at artifactory + "libs-snapshot"
 
 libraryDependencies ++= Seq(
-  "com.fasterxml.jackson.core"     % "jackson-annotations" % "2.8.10",
-  "com.fasterxml.jackson.core"     % "jackson-databind"    % "2.8.10",
-  "com.fasterxml.jackson.core"     % "jackson-core"        % "2.8.10",
-  "org.apache.commons" % "commons-compress" % "1.4.1",  // upgrading a transitive dependency to avoid security warnings
-  "org.apache.httpcomponents" % "httpclient" % "4.5.3",  // upgrading a transitive dependency to avoid security warnings
-  "ch.qos.logback" % "logback-classic" % "1.2.2",
+  "com.fasterxml.jackson.core"     % "jackson-annotations" % jacksonV,
```
aednichols: Builds and tests pass without these 3 Jackson lines

I think this might also eliminate an eviction warning, wasn't watching too closely

aednichols (commented):
Looking good... since we're already doing this, would like to get down to zero eviction warnings if possible

build.sbt
kyuksel: good catch..thanks!

build.sbt
```diff
@@ -5,10 +5,13 @@ name := "Agora"
 
 organization := "org.broadinstitute"
 
-scalaVersion := "2.11.12"
-dependencyOverrides += "org.scala-lang" % "scala-compiler" % "2.11.12"
+scalaVersion := "2.12.5"
```
davidangb: up for debate: should we skip 2.12.5 and stick with 2.12.4? https://contributors.scala-lang.org/t/scala-2-12-6-coming-soon/1729

davidangb:
once the other 2.12 PR merges and this is a nice clean rebased diff, I will scan it with SourceClear to assess potential vulnerabilities in dependency versions. I realize I didn't include this task in the original Jira ticket (and not everybody has a login) - let's coordinate.

build.sbt
geoffjentry: Fwiw we rolled back from 2.12.5 to 2.12.4

kyuksel: Thanks for bringing this up! I don't feel strongly either way. Will roll back to `2.12.4` unless I hear an argument for not to.

kyuksel:
> since we're already doing this, would like to get down to zero eviction warnings if possible

[This post](https://stackoverflow.com/a/42520384/2851999) summarizes my thoughts about eviction warnings in general. In our case, we're already at the highest versions possible (minus the Scala version) and downgrading doesn't feel right. Does that make sense?

aednichols (approved):
👍

PR #258, opened 2018-04-04
GAWB-2840: Improve WDL error messages while creating/copying methods

kyuksel:
- Addresses [GAWB-2840](https://broadinstitute.atlassian.net/browse/GAWB-2840): See below for snapshots of some of the error messages. The redundant printing of `ERROR` comes from `wdl4s` (or the UI, depending on how you look at it).
- Prefixes errors due to invalid WDLs provided during method creation/copying, with `Invalid WDL:`
- Makes `AgoraBusiness.checkValidPayload` safer by handling the case where `AgoraEntity.payload` may be empty
- Refactors `AgoraImportSpec` to remove duplicate code between method `create` and `copy` APIs, and groups the tests as `positive` and `negative`


aednichols:
Can you give some example before and after error messages?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -103,49 +106,55 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
-          }
-
-        case _ => //hello "shouldn't get here" my old friend
-          Failure(ValidationException(s"AgoraEntity $agoraEntity has no type!"))
+    def withPayload(payload: String)(f: String => Try[Any]): ReadWriteAction[T] = {
+      f(payload) match {
+        case Success(_) => op
+        case Failure(e: SyntaxError) =>
+          DBIO.failed(ValidationException(s"$errorMessagePrefix ${e.getMessage}", e.getCause))
+        case Failure(e: WdlValidationException) =>
```
aednichols: Why are there two cases for this if they return the exact same message?

kyuksel:
> Can you give some example before and after error messages?

Certainly. The ones on the right-hand side are the newer versions.

import_blah

blah

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
kyuksel: Good catch! I've aggregated them.

aednichols (commented):
This looks good to me, but I can't really thumb until I can inspect the clean diff. This is nothing to do with your code, just the branch shenanigans forced by dev freeze

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -103,49 +106,55 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
-      case Success(_) => op
-      case Failure(e: SyntaxError) =>
-        DBIO.failed(ValidationException(e.getMessage, e.getCause))
-      case Failure(regret) =>
-        DBIO.failed(regret)
     }
+
+    agoraEntity.payload.map { payload =>
+      withPayload(payload) {
+        payload => agoraEntity.entityType match {
```
davidangb: hmm ... I really don't like `payload.map{ payload => withPayload(payload) { payload =>` ... that's 5 instances of the word "payload" in a three-deep nesting. Additionally, I find it awkward that `op` is called from an inner helper.

I **do** like abstracting different parts of the logic into helper functions, but I think this abstraction actually adds complexity. I honestly liked the original two-step architecture better:
1. Given a payload and entityType, generate a `Try`
2. Given a `Try`, generate a `DBIOAction` (potentially by calling `op`)

I think the original **code** could be cleaned up but I like the original **architecture** better.

kyuksel: Thanks for the feedback! My main goal was to guard against the possibility of having a `None` `payload` and if so, generate a useful error message. Let me know if the newest version accomplishes that better. The diff from the original version is basically:

```scala
agoraEntity.payload match {
  case None =>
    DBIO.failed(ValidationException(s"Agora entity $agoraEntity has no payload."))
  case Some(payload) =>
    business as usual ...
```

kyuksel:
jenkins retest

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -102,50 +105,53 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
-    payloadOK match {
-      case Success(_) => op
-      case Failure(e: SyntaxError) =>
-        DBIO.failed(ValidationException(e.getMessage, e.getCause))
-      case Failure(regret) =>
-        DBIO.failed(regret)
+        payloadOK match {
+          case Success(_) => op
+          case Failure(e @ (_: SyntaxError | _: WdlValidationException)) =>
+            DBIO.failed(ValidationException(s"$errorMessagePrefix ${e.getMessage}", e.getCause))
```
aednichols, line 143: `errorMessagePrefix` is used exactly once, IMO putting it in a variable is not worth it (and adds a step when locating the source of the error)

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -102,50 +105,53 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
     }
   }
 
-  private def checkValidPayload[T](agoraEntity: AgoraEntity, username: String)(op: => ReadWriteAction[T]): ReadWriteAction[T] = {
```
aednichols, line 105: Makes sense to drop the parameter, though sometimes we choose to keep 'em around to make it very obvious that sibling methods operate on similar data (I have no position on this particular one)

aednichols (approved):
👍 the changes are nice-to-haves

davidangb (approved):
:+1:

PR #259, opened 2018-05-29
adopt cromwell32 jar

davidangb:
DataBiosphere/firecloud-app#76

adopt the new cromwell32 jar, including an `import` change for compatibility.


jmthibault79 (approved):
:+1: pending perf tests

PR #274, opened 2019-03-22
implement pet SA switcharoo

dvoet:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
@@ -67,6 +67,9 @@ object AgoraConfig {
   lazy val gcsUserEmail = config.as[String]("gcs.user.email")
   lazy val adminGoogleGroup = config.as[Option[String]]("admin.google.group")
 
+  // sam
+  lazy val samUrl = config.as[String]("sam.url")
+
   //Config Settings
   object SwaggerConfig {
     private val swagger = config.getConfig("swagger")
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/SamClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/SamClient.scala
@@ -1,0 +1,25 @@
+package org.broadinstitute.dsde.agora.server.dataaccess
+
+import org.broadinstitute.dsde.agora.server.AgoraConfig
+import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi
+import org.broadinstitute.dsde.workbench.client.sam.{ApiClient, ApiException}
+
+import scala.util.{Failure, Success, Try}
+
+object SamClient {
+  private def samUserApi(accessToken: String) = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(AgoraConfig.samUrl)
+    new UsersApi(apiClient)
+  }
+
+  def getUserEmail(token: String): Option[String] = {
+    val api = samUserApi(token)
+    Try(api.getUserStatusInfo()) match {
+      case Success(userInfo) => Option(userInfo.getUserEmail)
+      case Failure(e: ApiException) if e.getCode == 404 => None
+      case Failure(regrets) => throw regrets
+    }
+  }
+}
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/SamClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/SamClient.scala
@@ -1,10 +1,14 @@
 package org.broadinstitute.dsde.agora.server.dataaccess
 
+import java.util
+
+import akka.http.scaladsl.model.StatusCodes
 import org.broadinstitute.dsde.agora.server.AgoraConfig
 import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi
-import org.broadinstitute.dsde.workbench.client.sam.{ApiClient, ApiException}
+import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo
+import org.broadinstitute.dsde.workbench.client.sam.{ApiCallback, ApiClient, ApiException}
 
-import scala.util.{Failure, Success, Try}
+import scala.concurrent.{Future, Promise}
 
 object SamClient {
   private def samUserApi(accessToken: String) = {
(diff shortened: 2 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
```diff
@@ -2,25 +2,36 @@ package org.broadinstitute.dsde.agora.server.webservice.routes
...
  */
 class OpenIdConnectDirectives extends AgoraDirectives {
+  private val serviceAccountDomain = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
+
+  private def isServiceAccount(email: String) = {
+    serviceAccountDomain.pattern.matcher(email).matches
+  }
 
   override def commonNameFromRequest(magnet: ImplicitMagnet[ExecutionContext]): Directive1[String] = {
     headerValueByName("OIDC_CLAIM_name")
```
kcibul: In the case of the switcharoo, what should the name be?

dvoet: I don't know but this is unused anyway so I will delete

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -67,6 +67,9 @@ object AgoraConfig {
   lazy val gcsUserEmail = config.as[String]("gcs.user.email")
   lazy val adminGoogleGroup = config.as[Option[String]]("admin.google.group")
 
+  // sam
+  lazy val samUrl = config.as[String]("sam.url")
+
```
kcibul, line 72: dumb question -- why do we tend to make these things lazy?  what's the advantage?

dvoet: I was just following the pattern of the rest of the code in this file... I don't think there is much advantage here.

But there might be 3 reasons: 1) don't spend the time to load it if you never need it, 2) spread out performance impact (i.e. not all of these things are loaded at object instantiation) and 3) complicated timing - the code is simpler to have a val in this scope but it is not ready to be evaluated until later.

dvoet: actually, if it were not lazy then the fiab tests would probably fail because firecloud-develop does not yet have the config changes on the dev branch

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
```diff
@@ -2,25 +2,32 @@ package org.broadinstitute.dsde.agora.server.webservice.routes
...
 
-  override def commonNameFromRequest(magnet: ImplicitMagnet[ExecutionContext]): Directive1[String] = {
-    headerValueByName("OIDC_CLAIM_name")
+  private def isServiceAccount(email: String) = {
+    serviceAccountDomain.pattern.matcher(email).matches
   }
 
   override def usernameFromRequest(magnet: ImplicitMagnet[ExecutionContext]): Directive1[String] = {
-    headerValueByName("OIDC_CLAIM_email")
+    (headerValueByName("OIDC_CLAIM_email") & headerValueByName("OIDC_access_token")).tmap { case (email, token) =>
```
breilly2: Will `OIDC_access_token` always be present? `headerValueByName` will result in a `MissingHeaderRejection` if the header is missing. This is the first time Agora is expecting this header.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
```diff
@@ -2,25 +2,32 @@ package org.broadinstitute.dsde.agora.server.webservice.routes
...
+      if (isServiceAccount(email)) {
+        SamClient.getUserEmail(token).getOrElse(email)
```
breilly2: This is making a synchronous HTTP call, which is something new in this flow. `usernameFromRequest` is called pretty frequently, from many (all? almost all?) endpoints. Is it okay that this is synchronous? Are there performance considerations we should think about? I don't know what the volume of requests to Agora is. Should we consider caching the result of the call to Sam?

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
dvoet: it will always be present, the proxy does it and it is the same proxy as everyone else uses

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
dvoet: I don't want to cache until we have data to suggest there is a problem.

I agree with the async comment and made it so since this is within the routing.

breilly2: Do we have some confidence that it won't be an immediate problem or do we need to gather some data before putting this change into production? Do we know how frequently these endpoints are called, such as based on user activity browsing method repo, or based on the number of submissions, or based on the number of individual jobs? This could affect more than just Agora; it's asking Sam to do more work.

breilly2 (approved):
👍 with reservations. I agree with not adding something like sam response caching unless it's needed. I worry about putting this into production without thinking through the potential impact on Agora throughput/response time and Sam load.

src/main/scala/org/broadinstitute/dsde/agora/server/webservice/routes/OpenIdConnectDirectives.scala
dvoet: Over the highest volume 3 hour period in the last 7 days this api saw 23K requests. 13K responded in 15ms or less and almost all were under 100ms. The request rate from the rawls side is also throttled as there are only 10 workers responsible for submitting workflows to cromwell.

dvoet: it looks like the highest sustained workflow submission rate to cromwell over the past 90 days was about 90/minute

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
helgridly: i googled because i was interested and found [this](https://stackoverflow.com/questions/7484928/what-does-a-lazy-val-do) thread. tldr: what doge said, plus it can break cyclic dependencies

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/SamClient.scala
```diff
@@ -0,0 +1,44 @@
...
+  private def samUserApi(accessToken: String) = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(AgoraConfig.samUrl)
+    new UsersApi(apiClient)
+  }
+
+  def getUserEmail(token: String): Future[Option[String]] = {
+    val api = samUserApi(token)
+    val promise = Promise[Option[String]]()
```
helgridly, line 23: can you talk a bit about why we're using promises?

dvoet: https://docs.scala-lang.org/overviews/core/futures.html#promises

I don't want to use a Future because the sam client already is async. If I did
```
Future(api.getUserStatusInfoAsync(...))
```
Then I would be launching the call in the future, not waiting for it. What I really need is to do something when getUserStatusInfoAsync finishes. `Promise` provides a handy way to do that and convert the call into a `Future` which the rest of the code understands.

The sam client does have synchronous functions so I suppose I could also do
```
Future(api.getUserStatusInfo(...))
```
but then I need to think about what context the Future runs on. If it is on the main context then this using the `Future` does not really gain us anything since it still blocks threads that are not meant to be blocked. So I would need to make another context... and all of this would be silly anyway since the sam client is async under the hood and the sync call just has a wait in there.

helgridly: ah. i think my mistake here was assuming we wrote the client, and that the return type of `getUserStatusInfoAsync` was `Future`. neither of which is true (it's generated, and thus something insane). thanks for the splain

helgridly (approved):

PR #281, opened 2019-06-25
Agora WDL 1.0 Support

kcibul:
Addresses CA-207.  Specifically would like feedback on
 - WaasClient
 - Removing Import tests (see comment block in top of test file)

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


build.sbt
```diff
@@ -38,7 +38,9 @@ libraryDependencies ++= Seq(
   "com.typesafe.slick" %% "slick" % "3.2.2",
   "com.zaxxer" % "HikariCP" % "2.7.8",
   "mysql" % "mysql-connector-java" % "6.0.6",
-  "org.broadinstitute" %% "wdl-draft2" % cromwellV,
```
davidangb, line 41: `cromwellV` at line 14 is now unused

src/main/scala/org/broadinstitute/dsde/agora/server/AgoraConfig.scala
```diff
@@ -57,6 +58,10 @@ object AgoraConfig {
   lazy val mongoDbPassword = config.as[Option[String]]("mongodb.password")
   lazy val mongoDbDatabase = config.as[Option[String]]("mongodb.db").getOrElse("agora")
 
+  // Womtool-as-a-Service (Waas)
+  lazy val waasServer = config.as[Option[String]]("waas.server")
+  lazy val waasPath = config.as[Option[String]]("waas.path")
```
davidangb: is `wassPath`  used anywhere? If not, delete; if so, add a `path` entry in the reference.confs

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -105,19 +102,19 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
         val payloadOK = agoraEntity.entityType match {
           case Some(AgoraEntityType.Task) =>
-            WdlNamespace.loadUsingSource(payload, None, Option(Seq(httpResolver(_))))
+            WaasClient.validate(payload, accessToken)
           // NOTE: Still not validating existence of docker images.
           // namespace.tasks.foreach { validateDockerImage }
 
           case Some(AgoraEntityType.Workflow) =>
-            WdlNamespaceWithWorkflow.load(payload, Seq(httpResolver(_)))
+            WaasClient.validate(payload, accessToken)
```
davidangb: not your code, but could you collapse these first two `case`s to:
```
case Some(AgoraEntityType.Task) | Some(AgoraEntityType.Workflow) =>
```
to avoid the duplicated code?

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -380,14 +377,16 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource)(implicit ec: E
...
               // get the set of required input keys, optional input keys, and output keys for this WDL
-              val (optionalInputs, requiredInputs) = workflow.inputs.partition(_._2.optional)
-              val optionalInputKeys = optionalInputs.keySet.map(_.toString)
-              val requiredInputKeys = requiredInputs.keySet.map(_.toString)
-              val outputKeys:Set[String] = workflow.outputs.map(_.fullyQualifiedName.toString).toSet
+              // We expect fully qualified names, but WaaS doesn't return that. So we construct it here
+              val (optionalInputs, requiredInputs) = workflow.getInputs.asScala.partition(_.getOptional)
+              val optionalInputKeys = optionalInputs.map(workflow.getName + "." + _.getName).toSet
+              val requiredInputKeys = requiredInputs.map(workflow.getName + "." + _.getName).toSet
+              val outputKeys:Set[String] = workflow.getOutputs().asScala.map(workflow.getName + "." + _.getName).toSet
```
davidangb, line 384: a little syntactic sugar; your call whether or not you like this better for style:
```
	implicit val workflowName: String = workflow.getName
	def qualifyKeySet(keys: Iterable[String])(implicit workflowName: String): Set[String] = keys.map(k => s"$workflowName.$k").toSet

	// get the set of required input keys, optional input keys, and output keys for this WDL
	// We expect fully qualified names, but WaaS doesn't return that. So we construct it here
	val (optionalInputs, requiredInputs) = workflow.getInputs.asScala.partition(_.getOptional)
	val optionalInputKeys = qualifyKeySet(optionalInputs.map(_.getName))
	val requiredInputKeys = qualifyKeySet(requiredInputs.map(_.getName))
	val outputKeys = qualifyKeySet(workflow.getOutputs.asScala.map(_.getName))
```
definitely reduces the number of calls to `workflow.getName`, at cost of a bit more abstraction. This could be a lot cleaner if  `ToolInputParameter` and `ToolOutputParameter` had a shared base class.

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/WaasClient.scala
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  def describe(wdl: String, accessToken: String) :Try[WorkflowDescription] = {
+    Try {
+      womtoolApi(accessToken).describe("v1", wdl, null, null, null, null)
+    } recoverWith {
+      case e: ApiException => Failure(e)
+    }
+  }
```
davidangb, line 34: how about this for validate and describe?
```
  def describe(payload: String, accessToken: String) :Try[WorkflowDescription] =
    Try(womtoolApi(accessToken).describe("v1", payload, null, null, null, null))

  def validate(payload: String, accessToken: String): Try[String] = {
    describe(payload, accessToken) match {
      case Success(wd) if wd.getValidWorkflow => Success("OK")
      case Success(wd) =>
        // newline-delimited string of errors may not be user-friendly in the front end
        Failure(ValidationException(String.join("\n", wd.getErrors)))
      case Failure(ex) =>
        // do you want to log anything here? if not, could delete this case entirely
        Failure(ex)
    }
  }
```

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
```diff
@@ -101,32 +99,4 @@ object ModelSupport {
...
-  private def formatAuthorEmail(email: String, author: String): String = {
-    val stringFormat = "%s <%s>"
-    (email.trim.isEmpty, author.trim.isEmpty) match {
-      case (true, false) => author.trim
-      case (false, true) => email.trim
-      case (false, false) => stringFormat.format(author.trim, email.trim)
-      case _ => ""
-    }
-  }
-
```
davidangb, line 131: pull the author fixes into a separate PR and get that merged sooner?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
```diff
@@ -97,7 +98,93 @@ object AgoraTestData {
...
+                               |  ],
+                               |  "images": [],
+                               |  "submittedDescriptorType": {
+                               |    "descriptorType": "WDL",
+                               |    "descriptorTypeVersion": "draft-2"
+                               |  },
+                               |  "importedDescriptorTypes": [],
+                               |  "meta": {}
+                               |}""".stripMargin
+
```
davidangb: would it make more sense to save these payloads as json files in src/test/resources somewhere, and read them into lazy vals? Saves a bunch of annoying Scala code here.

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessIntegrationSpec.scala
```diff
@@ -9,50 +9,56 @@ import org.scalatest.{BeforeAndAfterAll, DoNotDiscover, FlatSpec}
...
   override protected def afterAll() = {
     clearDatabases()
+    stopMockWaas()
   }
 
   "Agora" should "be able to store a task configuration with a valid (extant) official docker image" in {
-    val entity = patiently(agoraBusiness.insert(testAgoraEntityWithValidOfficialDockerImageInWdl, mockAuthenticatedOwner.get)).addManagers(Seq(mockAuthenticatedOwner.get)).addIsPublic(false)
+    setSingleMockWaasDescribeOkResponse(genericDockerPayloadDescribeResponse)
+
+    val entity = patiently(agoraBusiness.insert(testAgoraEntityWithValidOfficialDockerImageInWdl, mockAuthenticatedOwner.get, mockAccessToken)).addManagers(Seq(mockAuthenticatedOwner.get)).addIsPublic(false)
```
davidangb, line 21: here - and in other tests - I'm worried about the usage of `setSingleMockWaasDescribeOkResponse`, which relies on mockserver's `Times.exactly(1)`. Since the mockserver instance is shared across all unit tests in a spec, and the  `Times.exactly(1)` is stateful, this means that:
* if we move to running the unit tests inside a spec in parallel, the mockserver responses are not threadsafe
* if an individual test fails/aborts before making its call to mockserver, the test will not consume its one-time call, and mockserver will return something unexpected to the next unit test

as a quick fix, you could use `beforeEach`/`afterEach` to set mockserver to a pristine state for this api, and add a comment that the spec should never be parallelized.

More ideal would be to set up the mockserver statelessly - have it inspect the payload or some other input and deterministically return success/failure based on what the caller sent (it doesn't have to actually validate; it can just do e.g. a substring lookup)

davidangb (commented):
comments inline!

src/test/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusinessIntegrationSpec.scala
kcibul: still looking at the other feedback (thanks!) but I wanted to comment on this one since I actually was doing something along these lines tonight before I saw your feedback!  I was annoyed by this as well, so managed to upgrade mockserver (was a pain) so I could use a substring expectation and set up nearly all the requests/responses globally.  A real WaaS would be deterministic (ie the same payload should return the same response) and this just parallels that.

kcibul: I just pushed that code BTW if you want to take a peek

src/main/scala/org/broadinstitute/dsde/agora/server/ga4gh/ModelSupport.scala
kcibul: I'm hoping to get this merged soon :D

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestData.scala
kcibul: Good idea -- I didn't do this for everything in that file b/c I think many of those existing tests should be re-examined, but I did it for everything used in this PR and set up the approach that should be easy to extend

src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
kcibul: Interesting approach -- I think it's 6 of one, half dozen of the other just like you point out.  I'm inclined to leave it since I think the existing one is a little more straightforward to understand

kcibul:
jenkins retest

kcibul:
jenkins retest

kcibul:
jenkins retest

kcibul:
jenkins retest

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/WaasClient.scala
davidangb: I definitely prefer the use of `Try()` over `try{...}`, and I'd like to see `validate` call `describe` instead of copy-pasting the `womtoolApi` invocation in both places.

src/test/resources/bad_wdl_version.describe.json
```diff
@@ -0,0 +1,13 @@
+{
+  "valid":false,
+  "errors":["ERROR: Finished parsing without consuming all tokens.\n\nversion plaid\n^\n     "],
```
davidangb: is the extra whitespace here - and in malformed_file.describe.json - relevant?

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestFixture.scala
```diff
@@ -25,6 +30,40 @@ trait AgoraTestFixture {
...
+    // necessary to declare this here in order to disambiguate overloaded method
+    val p2 : BodyWithContentType[String] = new StringBody(payload, true)
+    waasMockServer.when(
+      waasRequest
+        .withBody(p2)
+        )
+      .respond(
+        org.mockserver.model.HttpResponse.response().withBody(response).withStatusCode(OK.intValue).withHeader("Content-Type", "application/json"))
+
+  }
```
davidangb, line 61: nit: this isn't actually a substring check, is it? It checks for equality on the body.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraImportSpec.scala
```diff
@@ -9,14 +9,23 @@ import org.broadinstitute.dsde.agora.server.model.AgoraApiJsonSupport._
...
+//
+// NOTE: this test suite is ignored with the move to WaaS.  Previously WDL validation,
+// which includes resolving imports, was done in-process here and these tests were
+// checking that functionality.  However, at this point this is all being done now by WaaS
+// and testing of that functionality should be done there.  Also, since we are mocking out
+// WaaS for testing in Agora, nothing would be gained by having a test here with an import
+//
+// I would like to remove this test entirely, getting PR review feedback on that
+//
+@Ignore
```
davidangb: fine with me  to delete this entirely!

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
```diff
@@ -31,31 +32,39 @@ class AgoraMethodsSpec extends ApiServiceSpec with FlatSpecLike {
...
+    testEntity2WithId = patiently(agoraBusiness.insert(testEntity2, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntity3WithId = patiently(agoraBusiness.insert(testEntity3, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntity4WithId = patiently(agoraBusiness.insert(testEntity4, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntity5WithId = patiently(agoraBusiness.insert(testEntity5, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntity6WithId = patiently(agoraBusiness.insert(testEntity6, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntity7WithId = patiently(agoraBusiness.insert(testEntity7, mockAuthenticatedOwner.get, mockAccessToken))
+    testEntityToBeRedactedWithId = patiently(agoraBusiness.insert(testEntityToBeRedacted, mockAuthenticatedOwner.get, mockAccessToken))
+
+    stopMockWaas()
+    startMockWaas()
```
davidangb: this `beforeAll`  calls `startMockWaas()`, `stopMockWaas()`,  `startMockWaas()` ... reduce to one `startMockWaas()`?

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/CompatibleConfigurationIntegrationSpec.scala
```diff
@@ -30,80 +30,123 @@ class CompatibleConfigurationIntegrationSpec extends FreeSpec with ExecutionDire
   val methodsService = new MethodsService(permsDataSource) with ActorRefFactoryContext
   val testRoutes = ApiService.handleExceptionsAndRejections (methodsService.queryCompatibleConfigurationsRoute)
 
+  // The following methods create a method AgoraEntity while at the same time constructing the Mock Waas
+  // describe response.  This test is more difficult than the others because it generates many unique
+  // method payloads (WDLs) so it can test the compatibility of methods
```
davidangb: and, we define "compatibility" in this context as having matching inputs/outputs, which requires waas parsing!

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/CompatibleConfigurationIntegrationSpec.scala
```diff
@@ -30,80 +30,123 @@ class CompatibleConfigurationIntegrationSpec extends FreeSpec with ExecutionDire
...
+  private def generateMethodWithWaasResponse(label:String, inputs:Seq[String], outputs:Seq[String], optionalInputs:Seq[String] = Seq.empty[String]) = {
+    val method = testMethod(label, inputs, outputs, optionalInputs)
+    addSubstringResponse(method.payload.get, waasResponse(inputs, outputs, optionalInputs))
+    method
+  }
```
davidangb, line 41: I'm not gonna hold the PR for this, but for the sake of engineering goodness - this method is side-effecty. It returns the method payload (nice and stateless) but also writes to the mockserver instance (side effect). In an ideal world, this method would either be fully stateless: return a (method, waasResponse) pair that calling code inserts to the right places, OR fully action-oriented: insert the method into the db and set up the mockserver response.

Again, not worth refactoring, especially since this is test code, if it's a big deal and would delay the  PR.

davidangb (commented):
a few more comments inline - a few that need action. I'd especially like to see a little more cleanup on WaasClient validate/describe.

kcibul:
jenkins retest

src/test/scala/org/broadinstitute/dsde/agora/server/AgoraTestFixture.scala
kcibul: it's actually a substring.  the body of the post is actually a multi-part form response and so the body contains all the content separators, etc in addition to the payload.

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraImportSpec.scala
kcibul: poof!

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
kcibul: old code, removing

src/test/resources/bad_wdl_version.describe.json
kcibul: cleaned up

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/CompatibleConfigurationIntegrationSpec.scala
kcibul: added

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/WaasClient.scala
kcibul: 👍

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/CompatibleConfigurationIntegrationSpec.scala
kcibul: done!

kcibul:
jenkins retest

davidangb (approved):
:+1: nice!

One last change - with the updated version of mockserver and/or the new usages of mockserver, could you please add a `<logger name="org.mockserver.mock.HttpStateHandler" level="warn" ...` statement to logback.xml,  to keep unit test output clean? I don't need to re-review after that change.

dvoet (approved):

ansingh7115 (approved):

PR #288, opened 2019-10-21
Print out why status check failed [WA-45]

aednichols:
Right now I'm looking at an instance that reports
```
{"ok":false,"systems":{"Agora":{"ok":false,"messages":["Unable to gather subsystem status"]}}}
```
which is not very helpful in solving the problem.


rsasch (approved):
LGTM

kshakir (approved):

PR #298, opened 2019-12-03
Cache entity aliases [WA-77]

aednichols:
This changeset relocates an often-performed string concatenation behind a `lazy val`.

Profiling Agora under load to the `/definitions` endpoint showed that we were spending a surprising amount of CPU converting the `snapshotId` integer to string, for example.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
@@ -133,6 +133,12 @@ case class AgoraEntity(namespace: Option[String] = None,
                        managers: Seq[String] = Seq(),
                        public: Option[Boolean] = None) {
 
+  lazy val entityAlias: String =
+    namespace.get + "." + name.get + "." + snapshotId.get
+
+  lazy val namespaceAlias: String =
+    namespace.get
+
   def agoraUrl: String = {
     AgoraConfig.urlFromType(entityType) + namespace.get + "/" + name.get + "/" + snapshotId.get
   }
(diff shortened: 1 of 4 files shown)


kshakir (approved):

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
```diff
@@ -133,6 +133,12 @@ case class AgoraEntity(namespace: Option[String] = None,
                        managers: Seq[String] = Seq(),
                        public: Option[Boolean] = None) {
 
+  lazy val entityAlias: String =
+    namespace.get + "." + name.get + "." + snapshotId.get
```
salonishah11, line 137: `namespace.get` can be replaced by `namespaceAlias`

aednichols: That does work but I like the symmetry between the two functions and not having them rely on one another (not a strong opinion)

kshakir: If you do decide to leave it, how about a comment so someone doesn't come behind you and change it without seeing this PR?

salonishah11 (approved):

src/main/scala/org/broadinstitute/dsde/agora/server/model/AgoraEntity.scala
aednichols: I will accept the majority opinion in that case 🙂

PR #320, opened 2022-06-02
BW-1251: Automatically lowercase any email addresses returned via listX queries

cjllanwarne:


mspector (approved):
LGTM!

aednichols (commented):
I am slightly nervous about introducing a new uppercase != lowercase issue somewhere else.

e.g. the LHS comes from a different Agora call that still has uppercase, and the RHS comes from one of these calls that is lowercased.

The UI change seems strictly safe in the sense that it is accepting a broader range of inputs, and I think it's sufficient.

cjllanwarne:
@aednichols sounds reasonable. Assuming the problem is fixed in the next release we can just close this PR. I'll add a reminder to check in on that next week.

aednichols:
Sounds good and thank you in advance for checking!

cjllanwarne:
Closing, "won't do". Already fixed by the UI changes and this one is much more risky because it hits actual security infrastructure.

PR #324, opened 2022-08-22
BW-1197: Add Cache-Control no-cache header on responses

cjllanwarne:


rtitle (approved):
Didn't test but looks like it does the right thing

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
```diff
@@ -80,6 +82,9 @@ class AgoraMethodsSpec extends ApiServiceSpec with AnyFlatSpecLike {
         val entities = responseAs[Seq[AgoraEntity]]
         assert(entities.toSet == brief(Seq(testEntity3WithId, testEntity4WithId, testEntity5WithId, testEntity6WithId, testEntity7WithId)).toSet)
         assert(status == OK)
+        println(response.headers)
+        println(headers)
```
salonishah11: Are the `println`s intentional?

salonishah11 (approved):

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/AgoraMethodsSpec.scala
cjllanwarne: Oops, definitely not!

PR #330, opened 2023-08-31
[WM-2065] Disallow emails with non-latin characters

dillydally414:
Verified that examples such as `♥`, `bołądź`, and `ı` now throw errors via bee's firecloudorch swagger. May want someone with firecloud/prod suitability to help on Tuesday to confirm if any examples of these exist in prod already that we should be looking for alternative solutions for.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
@@ -209,7 +209,7 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource) extends LazyLo
               }
 
               entityToInsertAction flatMap { entityToInsert =>
-                val userAccess = new AccessControl(username, AgoraPermissions(All))
+                val userAccess = AccessControl(username, AgoraPermissions(All))
 
                 for {
                   entityWithId <-
--- a/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/AccessControl.scala
+++ b/src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/AccessControl.scala
@@ -7,7 +7,10 @@ object AccessControl {
   def apply(accessObject: Tuple2[String, Int]): AccessControl = {
     val user = accessObject._1
     val permission = AgoraPermissions(accessObject._2)
-    AccessControl(user, permission)
+    if ("""[a-zA-Z0-9-_\.@]*""".r.matches(user))
+      AccessControl(user, permission)
+    else
+      throw new IllegalArgumentException("User contains invalid characters.")
   }
 
   def fromParams(params: Map[String, String]): AccessControl = {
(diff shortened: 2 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/agora/server/business/AgoraBusiness.scala
```diff
@@ -209,7 +209,7 @@ class AgoraBusiness(permissionsDataSource: PermissionsDataSource) extends LazyLo
               }
 
               entityToInsertAction flatMap { entityToInsert =>
-                val userAccess = new AccessControl(username, AgoraPermissions(All))
+                val userAccess = AccessControl(username, AgoraPermissions(All))
```
dillydally414, line 212: Removing usages of `new` ensures we follow requirements and use the proper `apply` method

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/AccessControl.scala
```diff
@@ -23,6 +23,8 @@ object AccessControl {
   final val publicUser:String = "public"
 }
 
-case class AccessControl(user: String, roles: AgoraPermissions)
+case class AccessControl(user: String, roles: AgoraPermissions) {
+  require("""[a-zA-Z0-9-_\.@]*""".r.matches(user), "User contains invalid characters.")
```
dillydally414, line 27: Added matching for "valid" user emails

src/test/scala/org/broadinstitute/dsde/agora/server/webservice/PermissionIntegrationSpec.scala
```diff
@@ -182,14 +182,24 @@ class PermissionIntegrationSpec extends AnyFlatSpec with ScalatestRouteTest with
...
-      "/" + agoraEntity1.snapshotId.get + "/" + "permissions" + s"?user=$owner2&roles=All") ~>
+      "/" + agoraEntity1.snapshotId.get + "/" + "permissions" + s"?user=${owner2.get}&roles=All") ~>
       routes ~>
       check {
         assert(status == OK)
         assert(responseAs[String] contains "Manage")
       }
   }
 
+  "Agora" should "not allow authorized users to insert an entity permission with non-latin characters." in {
```
dillydally414, line 193: New test using url-escaped dotless i

src/main/scala/org/broadinstitute/dsde/agora/server/dataaccess/permissions/AccessControl.scala
kpierre13: Are chars such as Ø, allowed? Or was this difficult to implement?

dillydally414: Currently this is disallowed because it's not a simple latin letter. Without listing out specific cases I think it would be difficult to ex: allow `Ø` but disallow `ı`

kpierre13 (approved):

mspector (approved):

